PR #11, opened 2021-03-08
Add support for `__typename` fields on namespace objects

Ackar:
This adds support for `__typename` fields on namespace objects, e.g.:
```graphql
{
  myNamespace {
    __typename
  }
}
```
As these fields are quite special (they are not attached to a specific service) and we are not guaranteed to actually query a single service (see above example), I added some new logic for "bramble only" fields.

So:
- In the planning: allocate these fields to a special `__bramble` service
- In the execution: use `executeBrambleStep` to execute that special step

### DIFF
--- a/execution.go
+++ b/execution.go
@@ -157,13 +157,6 @@ func (s *ExecutableSchema) ExecuteQuery(ctx context.Context) *graphql.Response {
 			result[f.Alias] = s.resolveType(ctx, perms.FilterSchema(s.MergedSchema), &ast.Type{NamedType: name}, f.SelectionSet)
 		case "__schema":
 			result[f.Alias] = s.resolveSchema(ctx, perms.FilterSchema(s.MergedSchema), f.SelectionSet)
-		case "__typename":
-			if op.Operation == ast.Query {
-				result[f.Alias] = queryObjectName
-			}
-			if op.Operation == ast.Mutation {
-				result[f.Alias] = mutationObjectName
-			}
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


execution.go
```diff
@@ -566,6 +559,10 @@ func newQueryExecution(client *GraphQLClient, schema *ast.Schema, tracer opentra
 func (e *QueryExecution) execute(ctx context.Context, plan *queryPlan, resData map[string]interface{}) []*gqlerror.Error {
 	e.wg.Add(len(plan.RootSteps))
 	for _, step := range plan.RootSteps {
+		if step.ServiceURL == "__bramble" {
```
pkqk, line 562:
```suggestion
		if step.ServiceURL == internalServiceName {
```

worth extracting this to a `const` similar to the others in `schema.go`

execution.go
```diff
@@ -764,6 +761,30 @@ func (e *QueryExecution) executeChildStep(ctx context.Context, step *queryPlanSt
...
+// fields. This is used for namespaces as they do not belong to any service.
+func buildTypenameResponseMap(ss ast.SelectionSet, currentType string) map[string]interface{} {
+	res := make(map[string]interface{})
+	for _, f := range selectionSetToFields(ss) {
+		if len(f.SelectionSet) > 0 {
+			res[f.Alias] = buildTypenameResponseMap(f.SelectionSet, f.Definition.Type.Name())
+			continue
+		}
+
+		if f.Name == "__typename" {
```
pkqk, line 781:
```suggestion
		if f.Name == introspectionTypeNameField {
```
also `__schema` to `introspectionSchemaField`?

pkqk (approved):

execution.go
Ackar: I left this one as is because it's a standard field documented in the graphql spec. So the value shouldn't change and I think it makes the code easier to read. We also already use this kind of values elsewhere in the code.

PR #12, opened 2021-03-10
Configure GQLClient max-response-size

LeonHuston:
This allows you to specify in your initial config files a Client max-response-size for your GraphQL clients.
It will default to the old hard-coded value of 1MB (1024*1024)

### DIFF
--- a/client.go
+++ b/client.go
@@ -21,12 +21,12 @@ type GraphQLClient struct {
 	Tracer          opentracing.Tracer
 }
 
-func NewClient() *GraphQLClient {
+func NewClient(maxResponseSize int64) *GraphQLClient {
 	return &GraphQLClient{
 		HTTPClient: &http.Client{
 			Timeout: 5 * time.Second,
 		},
-		MaxResponseSize: 1024 * 1024,
+		MaxResponseSize: maxResponseSize,
 	}
 }
 
--- a/client.go
+++ b/client.go
@@ -21,12 +21,26 @@ type GraphQLClient struct {
 	Tracer          opentracing.Tracer
 }
 
-func NewClient(maxResponseSize int64) *GraphQLClient {
-	return &GraphQLClient{
+type ClientOpt func(*GraphQLClient)
+
+func NewClient(opts ...ClientOpt) *GraphQLClient {
+	c := &GraphQLClient{
 		HTTPClient: &http.Client{
 			Timeout: 5 * time.Second,
 		},
-		MaxResponseSize: maxResponseSize,
+		MaxResponseSize: 1024 * 1024,
+	}
+
+	for _, opt := range opts {
+		opt(c)
+	}
+
+	return c
+}
+
+func WithMaxResponseSize(maxResponseSize int64) ClientOpt {
+	return func(s *GraphQLClient) {
+		s.MaxResponseSize = maxResponseSize
 	}
 }
 
--- a/config.go
+++ b/config.go
@@ -233,7 +233,7 @@ func (c *Config) Init() error {
 		services = append(services, NewService(s))
 	}
 
-	client := NewClient(WithMaxResponseSize(c.MaxRequestsPerQuery))
+	client := NewClient(WithMaxResponseSize(c.MaxServiceResponseSize))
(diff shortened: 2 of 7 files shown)


docs/configuration.md
```diff
@@ -15,6 +15,7 @@ Sample configuration:
   "log-level": "info",
   "poll-interval": "5s",
   "max-requests-per-query": 50,
+  "max-client-response-size": ...
```
pkqk, line 18: github syntax highlighting doesn't like it when its invalid json :(
```suggestion
  "max-client-response-size": 1048576,
```

config.go
```diff
@@ -27,6 +27,7 @@ type Config struct {
 	PollInterval         string    `json:"poll-interval"`
 	PollIntervalDuration time.Duration
 	MaxRequestsPerQuery  int64 `json:"max-requests-per-query"`
+	MaxResponseSize      int64 `json:"max-client-response-size"`
```
pkqk, line 30: I don't think the naming is quite right here, you're configuring the max client response size of any internal clients created, not a limit on the response size bramble could return?

client.go
```diff
@@ -21,12 +21,12 @@ type GraphQLClient struct {
 	Tracer          opentracing.Tracer
 }
 
-func NewClient() *GraphQLClient {
+func NewClient(maxResponseSize int64) *GraphQLClient {
```
pkqk, line 24: I think an optional config object with an `MaxResponseSize` that defaults to `1MB` would be a cleaner interface, then we don't need to provide a default in tests and in the introspection file

execution.go
```diff
@@ -19,8 +19,8 @@ import (
 	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
-func newExecutableSchema(plugins []Plugin, maxRequestsPerQuery int64, services ...*Service) *ExecutableSchema {
-	client := NewClient()
+func newExecutableSchema(plugins []Plugin, maxRequestsPerQuery, maxResponseSize int64, services ...*Service) *ExecutableSchema {
```
pkqk, line 22: might be worth reconsidering the signature of this function? perhaps give it a client that is already configured rather than expanding the argument list to include options for a client.

pkqk (approved):

client.go
lucianjon: Yep, while you cannot always truely make backwards incompatible changes due to the signature needing to change. This change would break any existing code importing bramble and calling `bramble.NewClient()` 

I'd recommend something like 

```
type ClientOpt func(*GraphQLClient)

func NewClient(opts ...ClientOpt) *GraphQLClient {
	c := &GraphQLClient{
		HTTPClient: &http.Client{
			Timeout: 5 * time.Second,
		},
		maxResponseSize: 1024 * 1024,
	}
	for _, opt := range opts {
		opt(c)
	}
	return c
}

func WithMaxResponseSize(maxResponseSize int64) ClientOpt {
	return func(s *GraphQLClient) {
		s.maxResponseSize = maxResponseSize
	}
}
```

execution.go
lucianjon: agreed, you should try avoid passing arguments to a function just so it can pass those along. Like Adam said, passing the already configured client in would be more clear.

pkqk (approved):

config.go
```diff
@@ -27,6 +27,7 @@ type Config struct {
 	PollInterval         string    `json:"poll-interval"`
 	PollIntervalDuration time.Duration
 	MaxRequestsPerQuery  int64 `json:"max-requests-per-query"`
+	MaxResponseSize      int64 `json:"max-service-response-size"`
```
Ackar, line 30: The name should probably be the same between the struct and the json tag (`MaxServiceResponseSize`)

config.go
```diff
@@ -231,7 +233,8 @@ func (c *Config) Init() error {
 		services = append(services, NewService(s))
 	}
 
-	es := newExecutableSchema(c.plugins, c.MaxRequestsPerQuery, services...)
+	client := NewClient(WithMaxResponseSize(c.MaxRequestsPerQuery))
```
Ackar, line 236: Using the wrong config here

Ackar (approved):

PR #24, opened 2021-04-06
Remove naming convention checks

### LINKED ISSUE #21: Strict validation for enum types
First of all, thanks for making this! Our organisation is using apollo for schema stitching since a long time, and we were looking at alternatives to it. This looks really promising 😀. 

I was trying this out with one micro-service, and ran into this error:
```enum value isn't ALL_CAPS```

I understand that enum values _ideally_ should be all caps, but it is not strictly enforced by the graphql specification. Is this restriction really necessary ?

Ackar:
Resolves #21.


pkqk (approved):

nmaquet:
Here are some comments on this for future reference. This extra validation made sense when Bramble was a proprietary Movio project. Now that Bramble is open source, it no longer makes sense to impose GraphQL naming best practices on all users. We do still recommend following GraphQL naming best practices but Bramble won't stop you from using any compliant schema.

PR #28, opened 2021-04-13
Fix id field being incorrectly added to some queries

Ackar:
The id field was being included in the query even when not required, i.e. even when there is a single query step and the data is not going to be merged with another step.
And because in that case the result just gets forwarded as is, the field was not getting removed during formatting.

This fixes query planning to not add the id field when the data won't be merged, or in other words:
- The step is not a child step; AND
- The step doesn't have any child step

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -60,7 +60,7 @@ func Plan(ctx *PlanningContext) (*queryPlan, error) {
 		return nil, fmt.Errorf("not implemented")
 	}
 
-	steps, err := createSteps(ctx, nil, parentType, "", ctx.Operation.SelectionSet)
+	steps, err := createSteps(ctx, nil, parentType, "", ctx.Operation.SelectionSet, false)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


plan.go
```diff
@@ -69,7 +69,7 @@ func Plan(ctx *PlanningContext) (*queryPlan, error) {
 	}, nil
 }
 
-func createSteps(ctx *PlanningContext, insertionPoint []string, parentType, parentLocation string, selectionSet ast.SelectionSet) ([]*queryPlanStep, error) {
+func createSteps(ctx *PlanningContext, insertionPoint []string, parentType, parentLocation string, selectionSet ast.SelectionSet, substep bool) ([]*queryPlanStep, error) {
```
pkqk, line 72:
```suggestion
func createSteps(ctx *PlanningContext, insertionPoint []string, parentType, parentLocation string, selectionSet ast.SelectionSet, childstep bool) ([]*queryPlanStep, error) {
```
Use the same name as the other function?

plan.go
```diff
@@ -78,7 +78,7 @@ func createSteps(ctx *PlanningContext, insertionPoint []string, parentType, pare
 	}
 
 	for location, selectionSet := range routedSelectionSet {
-		selectionSetForLocation, childrenSteps, err := extractSelectionSet(ctx, insertionPoint, parentType, selectionSet, location)
+		selectionSetForLocation, childrenSteps, err := extractSelectionSet(ctx, insertionPoint, parentType, selectionSet, location, substep)
```
pkqk, line 81:
```suggestion
		selectionSetForLocation, childrenSteps, err := extractSelectionSet(ctx, insertionPoint, parentType, selectionSet, location, childstep)
```

pkqk (approved):

PR #32, opened 2021-05-05
Fix inline fragment merge

Ackar:
When dealing with fragment spreads we usually want to pass the response from the client as is, because we don't necessarily know what type was matched. In order to do that we need to not unmarshal the response more than necessary.

So when merging results maps, unmarshal into a `map[string]json.RawMessage`  instead of `map[string]interface{}`, this way we recursively unmarshal only what's strictly necessary to do the merging.

### DIFF
--- a/execution.go
+++ b/execution.go
@@ -892,10 +892,12 @@ func mergeMaps(dst, src map[string]interface{}) {
 
 			switch value := v.(type) {
 			case json.RawMessage:
-				var m map[string]interface{}
+				// we want to unmarshal only what's necessary, so unmarshal only
+				// one level of the result
+				var m map[string]json.RawMessage
 				_ = json.Unmarshal([]byte(value), &m)
-				dst[k] = m
-				aValue = m
+				aValue = jsonMapToInterfaceMap(m)
+				dst[k] = aValue
 			case map[string]interface{}:
 				aValue = value
 			default:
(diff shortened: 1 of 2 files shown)


execution.go
```diff
@@ -183,6 +183,9 @@ func (s *ExecutableSchema) ExecuteQuery(ctx context.Context) *graphql.Response {
 		return graphql.ErrorResponse(ctx, err.Error())
 	}
 
+	b, _ := json.MarshalIndent(plan, "", "  ")
+	fmt.Println(string(b))
```
pkqk, line 187: Left the prinln in

execution.go
```diff
@@ -892,10 +895,12 @@ func mergeMaps(dst, src map[string]interface{}) {
 
 			switch value := v.(type) {
 			case json.RawMessage:
-				var m map[string]interface{}
+				// we want to unmarshal only what's necessary, so unmarshal only
```
pkqk, line 898: Is the `RawMessage` always going to be an Object? It could also be an Array couldn't it

pkqk (approved):

execution.go
Ackar: 🤦

execution.go
Ackar: This function is only called when merging root queries together and I can't think of a case where we'd need to unmarshal an array.
If we need to merge arrays it will be for boundary types and thus will be done in `executeChildStep`.
So I think this one should be safe.

PR #33, opened 2021-05-11
Fix missing recursive call filterDefinition

Ackar:
Fixed an issue during schema filtering using permissions.
This issue only occurred when using interfaces referencing interfaces referencing object types.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -162,6 +162,7 @@ func filterDefinition(sourceSchema *ast.Schema, visited map[string]bool, types m
 			if typ.Kind == ast.Interface {
 				for _, pt := range sourceSchema.PossibleTypes[typ.Name] {
 					types[pt.Name] = pt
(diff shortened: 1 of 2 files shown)


auth.go
```diff
@@ -162,6 +162,7 @@ func filterDefinition(sourceSchema *ast.Schema, visited map[string]bool, types m
 			if typ.Kind == ast.Interface {
 				for _, pt := range sourceSchema.PossibleTypes[typ.Name] {
 					types[pt.Name] = pt
+					_ = filterDefinition(sourceSchema, visited, types, pt, AllowedFields{AllowAll: true})
```
pkqk, line 165: how come the result is ok to throw away?

Ackar: The result of this function is a filtered definition for the sub type, but in that particular instance we are in the `AllowAll == true` branch, so we won't need that filtered definition as we're allowing everything.

Ackar: The result map is `types` so we add to this one, `visited` is just to make sure we don't create an infinite recursion.

PR #38, opened 2021-05-20
Include user agent in downstream queries

haydenwoodhead:
Adds user agent in the format `Bramble/$version ($operation)` when performing queries. E.g when querying services for an updated schema we would see `Bramble/v1.16.0 (update)`.

### DIFF
--- a/introspection.go
+++ b/introspection.go
@@ -16,12 +16,20 @@ type Service struct {
 	SchemaSource string
 	Schema       *ast.Schema
 	Status       string
+
+	client *GraphQLClient
 }
 
-func NewService(serviceURL string) *Service {
-	return &Service{
+func NewService(serviceURL string, client *GraphQLClient) *Service {
+	s := &Service{
 		ServiceURL: serviceURL,
 	}
+	if client == nil {
+		s.client = NewClient()
+	} else {
+		s.client = client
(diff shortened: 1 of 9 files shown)


introspection.go
```diff
@@ -16,12 +16,20 @@ type Service struct {
 	SchemaSource string
 	Schema       *ast.Schema
 	Status       string
+
+	client *GraphQLClient
 }
 
-func NewService(serviceURL string) *Service {
-	return &Service{
+func NewService(serviceURL string, client *GraphQLClient) *Service {
```
Ackar, line 23: I'm wondering if we should maybe do the client initialization here in `NewService`, this way we could have `NewClient(WithUserAgent(generateBrambleUserAgent("update")))` in only one place.
We'll lose the ability to inject a different client but I don't think we currently need that.

What do you think?

Ackar (commented):
Looks good to me. Could you maybe add a test for that?

introspection.go
haydenwoodhead: Yep that sounds good to me. I wasn't sure how reusable service needed to be so didn't want to prescribe only one user agent but if it's ok I'll switch it.

Ackar (approved):

PR #39, opened 2021-05-23
Make version public

haydenwoodhead:
Will allow custom builds to set and use the version


Ackar:
I think this should also be updated in the `ldflags`?

haydenwoodhead:
> I think this should also be updated in the ldflags?

Yes. Nice catch

Ackar (approved):

PR #42, opened 2021-06-16
fix go build command in README

djameson42:
go build command in README may need updating. 

looks to be introduced in #19


nmaquet:
Good catch, thanks for the PR :)

PR #46, opened 2021-06-27
Fixing a few setup issues

robrohanEP:
Was going to do a demo of bramble but ran into a few issues getting it running locally. This seems to fix it (mostly around the playground plugin crashing the service)


Ackar (approved):

Ackar:
Thanks @robrohanEP!

PR #54, opened 2021-07-13
Allow both array and singular boundary type getters

haydenwoodhead:
This allows us to have a schema similar to the example and to prefer the array method when it exists. 

```
type Query {
  foos(ids: [ID!]!): [Foo!]! @boundary
  foo(id: ID!): Foo @boundary
  service: Service!
}
```

We could be smart about when we query the singular or array method but I'm not in this PR.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -336,29 +336,36 @@ func usesFieldsBoundaryDirective(schema *ast.Schema) bool {
 
 // validateBoundaryFields checks that all boundary types have a getter and all getters are matching with a boundary type
 func validateBoundaryFields(schema *ast.Schema) error {
-	boundaryTypes := make(map[string]struct{})
+	boundaryTypes := make(map[string]bool)
 	for _, t := range schema.Types {
 		if t.Kind == ast.Object && isBoundaryObject(t) {
-			boundaryTypes[t.Name] = struct{}{}
+			boundaryTypes[t.Name] = false
 		}
 	}
 
 	for _, f := range schema.Query.Fields {
 		if hasBoundaryDirective(f) {
-			if _, ok := boundaryTypes[f.Type.Name()]; !ok {
+			hasBoundaryType, ok := boundaryTypes[f.Type.Name()]
+			if !ok {
 				return fmt.Errorf("declared boundary query for non-boundary type %q", f.Type.Name())
 			}
 
-			delete(boundaryTypes, f.Type.Name())
+			if hasBoundaryType {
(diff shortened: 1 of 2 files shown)


validate.go
```diff
@@ -403,7 +403,7 @@ func validateBoundaryQuery(f *ast.FieldDefinition) error {
 
 	if f.Arguments[0].Type.Elem != nil {
 		// array type check
-		if idsField := f.Arguments.ForName("ids"); idsField == nil || idsField.Type.String() != "[ID!]" {
+		if idsField := f.Arguments.ForName("ids"); idsField == nil || idsField.Type.String() != "[ID!]" && idsField.Type.String() != "[ID!]!" {
 			return fmt.Errorf(`boundary query must have a single "id: ID!" argument`)
```
pkqk, line 433:
```
return fmt.Errorf(`boundary query must have a single "id: ID!" argument`)
```
The error message should include both schema options:
> boundary query must have a single "id: ID!" or "ids: [ID!]!" argument

validate.go
```diff
@@ -403,7 +403,7 @@ func validateBoundaryQuery(f *ast.FieldDefinition) error {
 
 	if f.Arguments[0].Type.Elem != nil {
 		// array type check
-		if idsField := f.Arguments.ForName("ids"); idsField == nil || idsField.Type.String() != "[ID!]" {
+		if idsField := f.Arguments.ForName("ids"); idsField == nil || idsField.Type.String() != "[ID!]" && idsField.Type.String() != "[ID!]!" {
```
pkqk, line 432: We should only accept `[ID!]!`, I can't see a reason to accept a `null` value.

validate.go
```diff
@@ -336,29 +336,55 @@ func usesFieldsBoundaryDirective(schema *ast.Schema) bool {
 
 // validateBoundaryFields checks that all boundary types have a getter and all getters are matching with a boundary type
 func validateBoundaryFields(schema *ast.Schema) error {
-	boundaryTypes := make(map[string]struct{})
+	type boundaryType struct {
+		hasIndividualGetter bool
+		hasArrayGetter      bool
```
pkqk, line 341: The graphql spec term is a `List` not an `Array` so we should be consistent with the type name here and throughout the code too.

So error messages, variable names should mention list instead.

haydenwoodhead:
Going to close this pr. It doesn't make sense anymore but will reopen with some of the fixes

PR #55, opened 2021-07-15
Give better errors if duplicate boundary query exists

haydenwoodhead:


pkqk (approved):
We should possibly be referring to the _queries_ as _resolvers_ as that is the GraphQL nomenclature.

For example:
```
missing boundary queries for the following types
```
should be:
```
missing boundary resolver for the following types
```

PR #56, opened 2021-07-20
Fix incorrect json unmarshalling during execution

nmaquet:
This PR fixes a bug where json.Unmarshal was failing due to an incorrect assumption and the error wasn't checked.

### DIFF
--- a/execution.go
+++ b/execution.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"reflect"
 	"runtime/debug"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


execution.go
```diff
@@ -970,47 +973,32 @@ type insertionTarget struct {
...
 		}
-		return in
+		return parsedInput
 	case nil:
 		return nil
 	default:
-		panic(fmt.Sprintf("unhandled type: %s", reflect.TypeOf(in).Name()))
+		log.Panicf("unhandled type: %T", parsedInput)
 	}
+	return nil // unreachable
```
pkqk, line 1001: Might as well pull the `default` case out of the switch statement and put it's panic here, then there isn't an unreachable branch.

nmaquet: You'll still have the same issue I think. The go compiler doesn't recognise `log.Panicf` as a panic.

pkqk: oh, I wasn't considering the compilers feelings, just that then you don't need to say that part is unreachable

pkqk:
Closing this as the new execution work will replace it

PR #58, opened 2021-07-27
Preserve and use type information when marshalling fragment spreads

lucianjon:
When marshalling the result of a federated query that contained a fragment spread we would hit an error in the marshalling logic. 

The added tests show the conditions needed to trigger the problem but the gist is that the information needed to know when inside the context of a fragment spread was being thrown out. This caused the marshalling logic to look up type information on the interface and not the target type of the spread.

The fix for now is to intercept the call that flattens out fragments and preserve the `TypeCondition` of the fragment. Longer term we might want to consider a refactor of the `marshalResult` function

### DIFF
--- a/execution_test.go
+++ b/execution_test.go
@@ -21,6 +21,128 @@ import (
 	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
+func TestFederatedQueryFragmentSpreads(t *testing.T) {
+	serviceA := testService{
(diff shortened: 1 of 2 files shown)


pkqk (approved):
It might be cleaner to remove the call to `selectionSetToFields`/`selectionSetToFieldsWithTypeCondition` and flatten the selection set in the same marshal loop.

PR #62, opened 2021-08-13
DOC: fixed typo on installation url.

hsblhsn:
Go does not accept any protocol on package address. 
Fixed this error on installation in this commit `malformed import path "https:/github.com/movio/bramble/cmd/bramble": invalid char ':'`


pkqk (approved):
Thanks for the catch @hsblhsn, looks like it was written from memory rather than checked in a shell.

Issue #65, opened 2021-08-27
Always listens to 0.0.0.0 and few other configuration issues.


hsblhsn:
Hi there!

Bramble provides a nice way to configure itself. But there has some issue.

- [ ] The config file is in `json`. And `json` is not config language friendly. `yaml`, `toml` is far better.
- [ ] No way to specify network interface to listen. It always listens to `0.0.0.0`.
- [ ] Accepts ports as `int`. It should accept the whole listen address (ex: `0.0.0.0:8082`, `192.0.0.120:8083`) as string.

PR #66, opened 2021-08-27
Add network interface fields to configs.

hsblhsn:
The listen address should be string. But current configuration file accepts (only) network ports as integer. And then converts it to listen address. `(":%s", port)`

In this commit I have added a way to specifically provide network interface to listen. 
I wanted to make it full string. Instead of accepting port as integer, it should accept the whole network interface like `localhost:8080` or `0.0.0.0:8080`.

**For backward compatibility**, I just added a way to specify network interface addresses. 
From now users can specify what interface they want Bramble to listen. But the interface and port have to be in different fields. (will probably get fixed in v2).

Solves #65

### DIFF
--- a/config.go
+++ b/config.go
@@ -22,6 +22,9 @@ type PluginConfig struct {
 
 // Config contains the gateway configuration
 type Config struct {
+	GatewayInterface       string    `json:"gateway-interface"`
+	MetricsInterface       string    `json:"metrics-interface"`
+	PrivateInterface       string    `json:"private-interface"`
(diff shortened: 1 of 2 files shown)


pkqk:
Thanks for your PR @hsblhsn, I adapted it slightly in #70 

We run our deployment of bramble only in k8s so have never needed to specify anything other than the local port to listen on, this should give a bit more flexibility while providing backwards compatibility.

Does it fit your use case?

PR #69, opened 2021-08-30
Fix documentation on jwt auth

seeday:
The code looks for the `public_role` role.
https://github.com/movio/bramble/blob/b5a9e78402f436a6e975ccdc6bdaf60e223a1c30/plugins/auth_jwt.go#L117

But the docs claim it looks for the role `public`.


pkqk (approved):
Thanks for spotting that @seeday

PR #70, opened 2021-09-07
Allow specifying the full listen address in config

pkqk:
Adapted from the changes suggested by @hsblhsn in #66 but I decided it was better to allow the newer config field to fully specify the listen addresses and fallback to just the port on the unspecified address if that was all that was provided.

### DIFF
--- a/config.go
+++ b/config.go
@@ -22,6 +22,9 @@ type PluginConfig struct {
 
 // Config contains the gateway configuration
 type Config struct {
+	GatewayInterface       string    `json:"gateway-interface"`
+	MetricsInterface       string    `json:"metrics-interface"`
+	PrivateInterface       string    `json:"private-interface"`
 	GatewayPort            int       `json:"gateway-port"`
 	MetricsPort            int       `json:"metrics-port"`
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -42,19 +45,33 @@ type Config struct {
 	linkedFiles      []string
 }
 
+func (c *Config) interfaceAddress(iface string, port int) string {
```
hsblhsn, line 48: The name `interfaceAddress` does not look good to me in this case. Also the parameter `iface`.

Since it takes hole address via the `iface` parameter,  it should not be named `iface`. `addr` would be better in this case.

Also `interfaceAddress` returns the whole address and takes an address and port as param. `listenAddress` or `getListenAddress` would be better I think. 

What do you think?

hsblhsn (changes requested):
Just a few issue with naming things.

config.go
pkqk: Yeah good points about the naming, I've gone for `addrOrPort` for the private helper method.

segfault88 (approved):
Makes sense as a migration path 👍

PR #74, opened 2021-09-26
Rewritten execution pipeline

lucianjon:
The execution pipeline has been rewritten to improve readability in some
areas, and fix the result marshalling bugs that were popping up. The
summary of the changes is:

  * execution results are now collected before the merging step. This
  means all results are stored in their own map and will not be mutated
  as other steps execute.
  * null bubbling as per the GraphQL spec is now done in it's own step.
  * when formatting the response, the selection set is now used as the
  source of truth. This is done to better accomodate queries using
  fragments that currently break in some scenarios. These cases should
  now be fixed.
  * there is no more partial unmarshalling, all results are unmarshalled
  into either `map[string]interface{}` or `[]interface`. This means there
  are no steps of the pipeline that get conditionally skipped, as is the case
  right now.

History has been squashed down to keep it clean, @nmaquet and @segfault88 also put a ton of work into this.

There are two breaking changes contained within this PR:
  * support for the `Node` interface method of boundary field resolution has been dropped. We decided to move to the simpler `@boundary` approach around the time of officially releasing Bramble. However the code was kept around for some time to give people time to migrate. Support for the [currently documented boundary field resolution](https://movio.github.io/bramble/#/sharing-types) has not changed.
  * opentracing/jaeger support has been dropped. Due to the setup complexity of jaeger, the plugin didn't get much traction and we'd consider it unmaintained. To avoid keeping unmaintained code around the opentracing support has been dropped.


execution.go
```diff
@@ -222,27 +207,73 @@ func (s *ExecutableSchema) ExecuteQuery(ctx context.Context) *graphql.Response {
...
+		graphql.RegisterExtension(ctx, name, value)
+	}
 
 	if len(errs) > 0 {
 		AddField(ctx, "errors", errs)
 	}
+
 	return &graphql.Response{
-		Data:   res,
+		Data:   []byte(formattedResponse),
```
pkqk, line 274: Is this going to create large copy calls when casting a `string` to `[]byte`? It doesn't seem that in `formatResponseData` the type is required to be a string, you could build up a `[]byte` buffer instead.

query_execution.go
```diff
@@ -0,0 +1,898 @@
...
+}
+
+func (q *queryExecution) executeRootStep(step *QueryPlanStep) error {
+	var document string
+	if step.ParentType == "Query" {
+		document = "query " + formatSelectionSet(q.ctx, q.schema, step.SelectionSet)
+	} else if step.ParentType == "Mutation" {
+		document = "mutation " + formatSelectionSet(q.ctx, q.schema, step.SelectionSet)
+	} else {
+		return errors.New("non mutation or query root step")
```
pkqk, line 109:
```suggestion
		return errors.New("expected mutation or query root step")
```

query_execution.go
```diff
@@ -0,0 +1,898 @@
...
+		case *ast.Field:
+			if seenFields[selection.Alias] {
+				continue
+			}
+			seenFields[selection.Alias] = true
+			filteredSelectionSet = append(filteredSelectionSet, selection)
+		case *ast.InlineFragment:
+			fragment := selection
+			if objectTypename == "" {
+				return nil, errors.New("__typename must have been injected when dealing with fragments, check the planner is doing the right thing")
```
pkqk, line 871:
```suggestion
				return nil, errors.New("__typename must not be blank")
```
Is the longer error meaningful when the branch is not in development? Perhaps check before it is passed to `unionAndTrimSelectionSet`?

query_execution.go
```diff
@@ -102,11 +102,11 @@ func (q *queryExecution) Execute(queryPlan *QueryPlan) ([]executionResult, gqler
...
 func (q *queryExecution) executeRootStep(step *QueryPlanStep) error {
 	var document string
-	if step.ParentType == "Query" {
-		document = "query " + formatSelectionSet(q.ctx, q.schema, step.SelectionSet)
-	} else if step.ParentType == "Mutation" {
-		document = "mutation " + formatSelectionSet(q.ctx, q.schema, step.SelectionSet)
-	} else {
+
+	switch operationType := step.ParentType; operationType {
+	case "Query", "Mutation":
```
pkqk, line 107: `queryObjectName` and `mutationObjectName` are defined in `schema.go` in case you want to use consts for those values

query_execution.go
```diff
@@ -0,0 +1,889 @@
...
+
+	data, err := q.executeBoundaryQuery(documents, step.ServiceURL, boundaryField)
+	if err != nil {
+		q.writeExecutionResult(step, data, err)
+		return nil
+	}
+
+	q.writeExecutionResult(step, data, nil)
+
+	if len(data) > 0 {
```
nmaquet, line 180: @lucianjon Is it OK to ignore this silently? Should we at least log a warning?

plan.go
```diff
@@ -363,35 +369,41 @@ func stringArraysEqual(a, b []string) bool {
...
 	}
 
-	m[serviceURL][typeName] = BoundaryQuery{Query: query, Array: array}
+	m[serviceURL][typeName] = BoundaryField{Field: field, Array: array}
 }
 
-// Query returns the boundary query for the given service and type
-func (m BoundaryQueriesMap) Query(serviceURL, typeName string) BoundaryQuery {
+// Query returns the boundary field for the given service and type
+func (m BoundaryFieldsMap) Field(serviceURL, typeName string) BoundaryField {
```
nmaquet, line 398: @lucianjon Is defaulting to `node` still OK here? I thought we wanted to deprecate it?

query_execution.go
```diff
@@ -0,0 +1,889 @@
...
+
+	var locs []gqlerror.Location
+	for _, f := range selectionSetToFields(step.SelectionSet) {
+		pos := f.GetPosition()
+		if pos == nil {
+			continue
+		}
+		locs = append(locs, gqlerror.Location{Line: pos.Line, Column: pos.Column})
+
+		// if the field has a subset it's part of the path
```
nmaquet, line 245: @lucianjon `s/subset/selection set`

query_execution.go
```diff
@@ -0,0 +1,889 @@
...
+	if err != nil {
+		q.writeExecutionResult(step, data, err)
+		return nil
+	}
+
+	q.writeExecutionResult(step, data, nil)
+
+	if len(data) > 0 {
+		for _, childStep := range step.Then {
+			boundaryResultInsertionPoint, err := trimInsertionPointForNestedBoundaryStep(data, childStep.InsertionPoint)
```
nmaquet, line 182: @lucianjon Instead of having the execution code do the trimming here, would it be possible for the planning code to provide the correct insertion point in the first place?

query_execution.go
```diff
@@ -0,0 +1,889 @@
...
+			},
+		})
+	}
+
+	return outputErrs
+}
+
+// The insertionPoint represents the level a piece of data should be inserted at, relative to the root of the root step's data.
+// However results from a boundary query only contain a portion of that tree. For example, you could
+// have insertionPoint: ["foo", "bar", "movies". "movie", "compTitles"], with the below example as the boundary result we're
```
nmaquet, line 287: @lucianjon `, "movies". "movie",` the dot should be a comma I think

query_execution.go
```diff
@@ -0,0 +1,889 @@
...
+			if field.Name != "__typename" {
+				return nil, fmt.Errorf("buildTypenameResponseMap: expected __typename")
+			}
+			result[field.Alias] = parentTypeName
+		}
+	}
+	return result, nil
+}
+
+func fragmentImplementsAbstractType(schema *ast.Schema, objectType, interfaceType string) bool {
```
nmaquet, line 358: @lucianjon `interfaceType` is a bit misleading here since it can also be a union

query_execution.go
```diff
@@ -0,0 +1,889 @@
...
+	}
+
+	var (
+		documents      []string
+		selectionIndex int
+	)
+	for _, batch := range batchBy(ids, batchSize) {
+		var selections []string
+		for _, id := range batch {
+			selection := fmt.Sprintf("%s: %s(id: %q) %s", nodeAlias(selectionIndex), parentTypeBoundaryField.Field, id, selectionSetQL)
```
nmaquet, line 445: @lucianjon Should we retire the use of `nodeAlias` since we want to deprecate `node` boundary fields?

nmaquet (changes requested):
Some epic work @lucianjon 👏

query_execution.go
lucianjon: As discussed offline, the insertion point is still correct in the context of merging the step's result into the overall response data. This case here is an exception as the result of a boundary query only contains a small piece of the entire response.

The previous code got around this as it shared a single map to store results in and would merge as it went, which meant the insertion point would always be a valid path inside the response data passed to a child step.

query_execution.go
```diff
@@ -0,0 +1,889 @@
...
+		q.group.Go(func() error {
+			return q.executeChildStep(childStep, boundaryIDs)
+		})
+	}
+	return nil
+}
+
+func (q *queryExecution) executeDocument(document string, serviceURL string, response interface{}) error {
+	req := NewRequest(document).
+		WithHeaders(GetOutgoingRequestHeadersFromContext(q.ctx))
```
pkqk, line 142: Is this helper function worth it just to inline one assignment? It hides the modification side effect.

plan.go
lucianjon: Will change it to return an error as hitting that case means something has gone wrong

query_execution.go
lucianjon: Will drop the function as the name is no longer relevant but I think it makes sense to keep the behaviour of using `_{id}` as the key when doing non-array boundary field lookups. These are still the `@boundary` directive based lookups rather than the node ones.

query_execution.go
lucianjon: Not a fan of mutating the request after initialisation regardless but I agree this doesn't really help in that area. However setting public fields directly on the struct after initialisation felt more wrong to me than using a helper function, ideally the `Headers` field wouldn't be public at all but it's a little hard to change now. I don't feel too strongly either way though, my preference would to completely redo the initialisation at some point.

PR #81, opened 2021-11-01
Fix: expand all possible abstract ids

gmac:
This fixes a bug with the notoriously tricky hint selections of abstract types. 

## The problem

Here's a schema...

```graphql
type Apple @boundary {
  id: ID!
  name: String!
}
type Banana @boundary {
  id: ID!
  name: String!
}
union Fruit = Apple | Banana

type Query {
  fruit: Fruit
}
```

Then here's a query:

```graphql
query {
  fruit {
    ...on Apple { name }
  }
}
```

This query works fine if the endpoint returns an `Apple`, but it fails if a `Banana` is returned:

```json
{
  "errors": [
    {
      "message": "boundaryIDFromMap: 'id' or '_id' not found"
    }
  ],
  "data": null
}
```

Why does a `Banana` fail? It's because ID hint selections are only being added to _user-defined boundary fragments_, i.e: `Apple`. In the above selection, the user hasn't explicitly made a `Banana` selection, so there's no ID hint added for a banana. That means that when a `Banana` gets is returned, the executor comes up with a boundary type that returned no ID.

## The fix

This updates the query planner to expand abstract types with ID selection hints for all possible boundary types. The above request now comes out looking like this:

```graphql
query {
  fruit {
    ... on Apple {
      _id: id
    }
    ... on Banana {
      _id: id
    }
    ... on Apple {
      name
    }
  }
}
```

Now, regardless of what the user has selected, we know that all possible boundary results will come back with the expected ID. [GraphQL Tools stitching](https://github.com/ardatan/graphql-tools) does something very similar. It's generally better to normalize the request than to make complicated assumptions about the result.

**Gross, there's now two `...on Apple` fragments, can we consolidate those?** Sure, we _could_ using something like [selectionSetHasFieldNamed](https://github.com/gmac/bramble/blob/eda96f845882405193f1bc4262f8b54b4d5d8626/plan.go#L353-L361) for fragments. However, I'm generally against adding loops for the sake of human cosmetics when GraphQL servers don't care if fragments or fields get duplicated in the selection. But if there's strong resistance to this, I'm happy to prettify it.

## Tests

Opening this now for initial review on the implementation... I'm still learning Go, so I'm a bit slow to write code and still have to figure out how to write/run tests.

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -220,21 +220,48 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
 		}
 	}
 
-	// We need to add the id field only if it's a boundary type and the result
-	// is going to be merged with another step (we have children steps or it's a
-	// child step).
-	if parentType != queryObjectName && parentType != mutationObjectName &&
-		ctx.IsBoundary[parentType] &&
-		ctx.Schema.Types[parentType].Fields.ForName("id") != nil &&
-		(childstep || len(childrenStepsResult) > 0) {
-		if !selectionSetHasFieldNamed(selectionSetResult, "id") {
-			id := &ast.Field{
-				Alias:      "_id",
-				Name:       "id",
-				Definition: ctx.Schema.Types[parentType].Fields.ForName("id"),
+	parentDef := ctx.Schema.Types[parentType]
+	// For abstract types, add an id fragment for all possible boundary
+	// implementations. This assures that abstract boundaries always return
+	// with an id, even if they didn't make a selection on the returned type.
+	if parentDef.IsAbstractType() {
+		for implementationName, abstractTypes := range ctx.Schema.Implements {
+			if !ctx.IsBoundary[implementationName] {
+				continue
 			}
-			selectionSetResult = append([]ast.Selection{id}, selectionSetResult...)
+			for _, abstractType := range abstractTypes {
+				if abstractType.Name != parentType {
+					continue
+				}
+				implementationType := ctx.Schema.Types[implementationName]
+				possibleId := &ast.InlineFragment{
+					TypeCondition: implementationName,
+					SelectionSet: []ast.Selection{
+						&ast.Field{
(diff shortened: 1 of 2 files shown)


plan.go
```diff
@@ -220,20 +220,51 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
...
-		ctx.IsBoundary[parentType] &&
-		ctx.Schema.Types[parentType].Fields.ForName("id") != nil &&
-		(childstep || len(childrenStepsResult) > 0) {
-		if !selectionSetHasFieldNamed(selectionSetResult, "id") {
-			id := &ast.Field{
-				Alias:      "_id",
-				Name:       "id",
-				Definition: ctx.Schema.Types[parentType].Fields.ForName("id"),
+	parentDef := ctx.Schema.Types[parentType]
+	if parentDef.Kind == ast.Interface || parentDef.Kind == ast.Union {
```
nmaquet, line 224: You can use `parentDef.IsAbstractType()` here instead

nmaquet:
@gmac This is excellent. Thanks a lot for this. I'm happy to write unit tests for this if you'd prefer. Should be a simple case of a adding the Apple/Banana example to `plan_test.go`. As to whether it's OK to have two fragments on the same type in the query, IMHO it's completely fine. The spec allows its and all GraphQL implementations can deal with (and merging it would be a pain) so I'm happy with this as is.

lucianjon:
Good catch and thanks for the fix @gmac. I agree with @nmaquet that the approach makes sense, bramble isn't optimised to create human readable queries. It's good to see some of the edge cases around fragments _not_ getting matched being discovered and fixed, we had a similar case with #76.

gmac:
Thanks @nmaquet & @lucianjon, tests have been added. Bramble is teaching me Go, and I'm really enjoying it! This PR is good to go on my end!

PR #82, opened 2021-11-01
Update dependencies and update go to 1.17

lucianjon:
Bump up to go 1.17 and run a `go mod tidy -go=1.17`. The dependency updates are mostly minor with the exception of gqlgen bringing a newer version of gqlparser with it. gqlparser 2.2.0 implemented directive support, which in turn found some invalid schemas in tests.


nmaquet:
@lucianjon Might want to update the README to mention 1.17 as a prerequisite instead of 1.16?

lucianjon:
> @lucianjon Might want to update the README to mention 1.17 as a prerequisite instead of 1.16?

Yep done, missed those.

Issue #73, opened 2021-09-16
Does Bramble support resolving one field from multiple federated services?


ihac:
Hi, as you mentioned in the doc:
> With the exception of namespaces and the id field in objects with the @boundary directive, every field in the merged schema is defined in exactly one federated service.

Does this mean any field (even it's a boundary object) should be resolved by only one service? Or in other words, does Bramble support resolving a boundary object with unlimited recursion?

Here is the use case for your reference:

``` graphql
## foo service

type Foo @boundary {
  id: ID!
  title: String!
  bar: Bar!
}

type Bar @boundary {
  id: ID!
}


## bar service

type Bar @boundary {
  id: ID!
  subtitle: String!
}
```

The schema above works well in Bramble, as we could resolve `Foo.bar` within the `bar service`.

However, it starts failing when we add a new field in `Bar`, which can only be resolved in the third service:

``` graphql
## foo service

type Foo @boundary {
  id: ID!
  title: String!
  bar: Bar!
}

type Bar @boundary {
  id: ID!
}


## bar service

type Bar @boundary {
  id: ID!
  subtitle: String!
  baz: Baz!
}

type Baz @boundary {
  id: ID!
}


## baz service

type Baz @boundary {
  id: ID!
  status: String!
}
```

The query fails when trying to resolve `Foo.bar.baz.status`: `Cannot query field status on type Baz`.

pkqk:
Hi @ihac 

This should work, bramble currently has no protections against query complexity other than a timeout on execution time, so recursing down to dependent services should be fine.

I'm assuming you've cut out the fields on `Query` for looking up the `@boundary` types in the example schemas.

Can you share these example services so we could run a test with them?

ihac:
Hi @pkqk here is the repo: https://github.com/ihac/graphql-poc-playground/tree/master/bramble-federation if you wanna run a test.

And below are some failed queries (we expected them to work though):

``` graphql
query {
  randomContentWithScoreCard { # resolved by content service.
    id
    title
    score { # resolved by social service
      teamAScore
      teamBScore
      ads { # should be resolved by ads service
        id
        message # cannot query field "message" on type "TinyAdsBanner"
      }
    }
  }
}
```

``` graphql
# Same issue as above
query {
  masthead_v2(count: 10) { # resolved by masthead service
    __typename
    contentItemsWithScoreCard { # resolved by content service
      id
      title
      score { #should be resolved by social service
        id
[...]

PR #85, opened 2021-11-02
Fix: fully traverse nested field selections while planning

gmac:
As described in https://github.com/movio/bramble/issues/73, deeply-nested selection paths that transitioned between services were broken. This was a fairly significant issue.

## The problem

While extracting selection sets, the `mergedWithExistingStep` check was [treating _all_ consolidated fields as leaf values](https://github.com/movio/bramble/compare/main...gmac:fix-child-step-merging?expand=1#diff-11ffedf6a81044546aa19c8009998208f5a7afec3cdff04ca4281117b5ec0da7L163). That means that when a composite field was encountered with its own nested selections, that entire selection set was merged into the current planning step without traversing its children to delegate them properly. This resulted in fields belonging to foreign services getting lumped into the present service selection, and the query failing with a GraphQL validation message:

```
Cannot query field <remoteField> on type <LocalType>
```

## The fix

This adjusts the flow of `extractSelectionSet` so that all remote fields are collected up front and then create steps once together later in the process. This eliminates the need for the previous step-merging pattern, and is computationally simpler to create steps all at once rather than creating them incrementally for each remote step in sequence.

Note that this refactor looks bigger than it really is: this simplification was able to remove some conditional nesting and thus decrease the indent on a few existing code blocks. 

## Tests

Tests have been added. There's one update to an existing test where a benign array order has changed due to the revised sequencing of extracting remote fields. 

Resolves https://github.com/movio/bramble/issues/73.

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -117,6 +117,7 @@ func createSteps(ctx *PlanningContext, insertionPoint []string, parentType, pare
 func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentType string, input ast.SelectionSet, location string, childstep bool) (ast.SelectionSet, []*QueryPlanStep, error) {
 	var selectionSetResult []ast.Selection
 	var childrenStepsResult []*QueryPlanStep
+	var remoteSelections []ast.Selection
 	for _, selection := range input {
 		switch selection := selection.(type) {
 		case *ast.Field:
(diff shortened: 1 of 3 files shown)


plan_test.go
```diff
@@ -118,18 +118,18 @@ func TestQueryPlanABA2(t *testing.T) {
 				"SelectionSet": "{ _id: id compTitles(limit: 42) { id compTitles(limit: 666) { id } } }",
 				"InsertionPoint": ["movies"],
 				"Then": [
-				  {
+					{
 					"ServiceURL": "A",
 					"ParentType": "Movie",
 					"SelectionSet": "{ _id: id title }",
-					"InsertionPoint": ["movies", "compTitles"],
+					"InsertionPoint": ["movies", "compTitles", "compTitles"],
```
gmac, line 125: This change to an existing test is functionally identical to its old assertion. The only thing that's changed here is the array order of the resulting steps, which is benign to the execution process.

plan.go
```diff
@@ -136,44 +137,29 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
...
-					if err != nil {
-						return nil, nil, err
-					}
-					newField.SelectionSet = selectionSet
-					selectionSetResult = append(selectionSetResult, &newField)
-					childrenStepsResult = append(childrenStepsResult, childrenSteps...)
-				}
+			if loc != location {
+				// field transitions to another service location
+				remoteSelections = append(remoteSelections, selection)
```
gmac, line 142: This is the only actual change to this conditional: transitions in service location now collect remote fields into an array and resolve them all at once later in the process, versus resolving each remote field individually here. There's a lot less churn resolving them together, and it avoids the situation that `mergedWithExistingStep` was previously trying to address.

gmac:
@nmaquet FYI – looks like the map enumeration I did in https://github.com/movio/bramble/pull/81 has introduced a spurious test failure with the print order of fragments. I'll noodle on fixing it.

lucianjon:
> @nmaquet FYI – looks like the map enumeration I did in #81 has introduced a spurious test failure with the print order of fragments. I'll noodle on fixing it.

@gmac You've potentially already figured it out but this is due to this line https://github.com/movio/bramble/blob/main/plan.go#L228, as the iteration ordering over maps is intentionally random in golang. Sorting the output feels wrong because we don't actually care about ordering other than for the test. At the same time due to the fact the selection set is a string here makes it a difficult to fix comparison.

It might pay to create something custom for that test. If we can parse the selection set string back into the AST format we could simply check for the presence of each fragment.

gmac:
@lucianjon
> It might pay to create something custom for that test. If we can parse the selection set string back into the AST format we could simply check for the presence of each fragment.

Good call! Yep, I learned this morning about randomized map orders... it was a surprise to me coming from the land of Ruby and JS where enumerable keys maintain definition order. Thanks for the info and the tip. Will have another PR for that fix in the next day or two.

lucianjon:
@gmac Great find, looks good from my perspective. The fix and added tests make sense, I think the collect and then create approach is cleaner too.

Issue #48, opened 2021-07-05
Add Fedaration gateway example


thangld322:
The document is lacking of examples. Can you add a simple working example of how to implement Federation gateway with @bramble? I can't manage to do it. Thank you!

pkqk:
Hi @thangld322 

There are three example services in the [examples directory](/movio/bramble/tree/main/examples) if you run bramble with the `config.json` file set to point to those services
```json
{
  "services": [
    "http://nodejs-service:8080/graphql",
    "http://gqlgen-service:8080/graphql",
    "http://graph-gophers:8080/graphql"
  ]
}
```

Then you can request an arbitrary `Foo` object with any `id` value and each service will expose a field on the object.

```graphql
query {
  foo(id: "random") {
    nodejs
    gqlgen
    graphGophers
}
```

It's not a very detailed example but it does show the main features of the `@boundary` directive working.

I could put together a docker compose file that would run them all together if you would find that helpful to get the examples running?

thangld322:
Thank @pkqk 
I'll try that out.
```
I could put together a docker compose file that would run them all together if you would find that helpful to get the examples running?
```
This would be very helpful!

pkqk:
@thangld322 I haven't gotten a docker example yet but you can run the examples with these commands
```shell
cd examples/gqlgen-service
gqlgen generate
cd ../..
PORT=8092 go run ./examples/gqlgen-service/ &
PORT=8091 go run ./examples/graph-gophers-service/ &
BRAMBLE_SERVICE_LIST="http://localhost:8091/query http://localhost:8092/query" go run ./cmd/bramble
```

Then you can query to bramble at `http://localhost:8082/query` with:
```graphql
query {
  randomFoo {
    id
    gqlgen
    graphGophers
  }
}
```

zlenderja:
@pkqk I get a long list of errors when I run `gqlgen generate`

```
validation failed: packages.Load: /Users/zlender/code/go/bramble/examples/gqlgen-service/generated.go:242:18: WithPathContext not declared by package graphql
/Users/user/code/go/bramble/examples/gqlgen-service/generated.go:242:47: NewPathWithField not declared by package graphql
/Users/user/code/go/bramble/examples/gqlgen-service/generated.go:257:18: WithPathContext not declared by package graphql
/Users/user/code/go/bramble/examples/gqlgen-service/generated.go:257:47: NewPathWithField not declared by package graphql
....
```

Any tips much appreciated.

[...]

PR #86, opened 2021-11-03
Example of gateway operating in a docker compose environment

suessflorian:
Closes https://github.com/movio/bramble/issues/48


examples/docker-compose.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - 8080
+  nodejs-server:
+    build:
+      context: nodejs-service
+    expose:
+      - 8080
+  gateway:
+    image: ghcr.io/movio/bramble
+    volumes:
+      - ${PWD}/example-config.json:/config.json
```
pkqk, line 21: Where does the `PWD` value come from? Are you using a `.env` file that's not commited?
```suggestion
      - ./example-config.json:/config.json
```

examples/gqlgen-service/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang
+
+WORKDIR /go/src/app
+
+COPY . .
+RUN go mod init gqlen-server
```
lucianjon, line 6: @suessflorian I'd make these submodules of bramble (github.com/movio/bramble/examples/gqlgen-server) and instantiate them outside of docker. Just so the behaviour and dependencies are the same when running inside of docker and out.

pkqk: That will make any dependencies in the example services be dependencies of `bramble`, even if they are not necessary?

pkqk: I would do the build outside of the docker image if possible, the cache interactions with doing the build in the image slow things down a lot if you are changing the code.

lucianjon: Isn't that the case now?  These examples are going up the tree to bramble's `go.mod` and so their dependencies get injected. This should do the reverse of that.

If I do this:

```
cd examples/gqlgen-service && go mod init github.com/movio/bramble/examples/gqlgen && go mod tidy
cd ../../ && go mod tidy
```

It drops dependencies from bramble's main `go.mod` file.

lucianjon: By the way, by currently I mean before this PR. I just think we should move the `go mod init` to outside docker and commit the `go.mod` and `go.sum` files.

examples/gqlgen-service/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang
+
+WORKDIR /go/src/app
+
+COPY . .
+
+RUN go get
```
lucianjon, line 7: @suessflorian Think you're missing a `go generate` somewhere here.

Issue #88, opened 2021-11-03
Spurious failure in tests


gmac:
Introduced by the random enumeration order of this line: https://github.com/movio/bramble/pull/81/files#diff-11ffedf6a81044546aa19c8009998208f5a7afec3cdff04ca4281117b5ec0da7R228

I introduced the bug, so I'm happy to own the fix. I should have a PR by end of week.

PR #89, opened 2021-11-04
Fix: spurious test failures caused by #81

gmac:
Fixes the inconsistent test failures introduced by #81. That PR tested selection sets generated from an unordered map enumeration; as such, the tests fail periodically when the unordered enumeration doesn't match the way the tests were written.

This introduces a new `CheckUnorderedRootFieldSelections` helper that accepts an array of selections and assures that they're each found within the planned result's root field, regardless of appearance order.

Resolves https://github.com/movio/bramble/issues/88.

### DIFF
--- a/plan_fixtures_test.go
+++ b/plan_fixtures_test.go
@@ -4,12 +4,15 @@ import (
 	"encoding/json"
 	"sort"
 	"testing"
+	"context"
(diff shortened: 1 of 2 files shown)


nmaquet:
@gmac This is perfect, thanks!

PR #91, opened 2021-11-05
Always add __typename to abstract types

gmac:
Preemptively fixing the same issue described in https://github.com/movio/bramble/pull/81, but with `__typename`...

## The (potential) problem

Given that the execution resolver uses `__typename` while assembling abstract types, we need to assure that a typename is always returned. However, the automated `__typename` hints are currently only added to _user-defined_ fragments, and those aren't guaranteed to be present when making abstract type selections (case of point: selecting base interface fields).

While I haven't actually made this scenario fail, I strongly suspect that it can be broken.

## The (preemptive) fix

This adds `__typename` to all abstract types. In the event that the user doesn't make a fragment selection on an abstract, the selection will still request typename for resolution purposes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@
 .vscode
 .history
 .idea
+.DS_Store
 coverage.out
--- a/plan.go
+++ b/plan.go
@@ -262,6 +262,11 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
 				break
 			}
 		}
+		selectionSetResult = append(selectionSetResult, &ast.Field{
+			Alias: "__typename",
+			Name: "__typename",
+			Definition: &ast.FieldDefinition{Name: "__typename", Type: ast.NamedType("String", nil)},
+		})
 	// Otherwise, add an id selection to boundary types where the result
 	// will be merged with another step (i.e.: has children or is a child step).
 	} else if parentType != queryObjectName &&
--- a/plan_test.go
+++ b/plan_test.go
@@ -496,6 +496,7 @@ func TestQueryPlanExpandAbstractTypesWithPossibleBoundaryIds(t *testing.T) {
 		"name",
 		"... on Lion { _id: id }",
 		"... on Snake { _id: id }",
+		"__typename",
 	}
 	PlanTestFixture3.CheckUnorderedRootFieldSelections(t, query, rootFieldSelections)
 }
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -2,4 +2,5 @@
 .vscode
 .history
 .idea
+.DS_Store
```
gmac, line 5: the bane of MacOS developers...

gmac:
@nmaquet – this is ready for consideration.

plan.go
```diff
@@ -262,6 +262,11 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
 				break
 			}
 		}
+		selectionSetResult = append(selectionSetResult, &ast.Field{
```
nmaquet, line 265: Could you add a `if !selectionSetHasFieldNamed(selectionSetResult, "__typename") { }` around the append? Our clients send `__typename` through everywhere so this will avoid a lot of duplicate `__typename` selections.

plan_test.go
```diff
@@ -638,7 +640,7 @@ func TestQueryPlanSupportsUnions(t *testing.T) {
         {
           "ServiceURL": "A",
           "ParentType": "Query",
-          "SelectionSet": "{ animals { ... on Dog { name __typename } ... on Cat { name __typename } ... on Snake { name __typename } } }",
```
nmaquet, line 641: Might be worth duplicating this test with `DontDuplicateTypename` as a suffix with the selection set `{ animals { ... on Dog { name __typename } ... on Cat { name __typename } ... on Snake { name __typename }  __typename } }` to show that `__typename` does not get duplicated.

nmaquet (changes requested):
Another nice find, thanks @gmac.

suessflorian:
I'm thinking; if we guarantee a `__typename` injection when the parent type presents itself as an abstract type, can we then omit the injection when encountering fragment spreads?

gmac:
@suessflorian — I had the same thought… I haven’t dug deep enough into the execution pipeline yet for the answer though. The wildcard would be if typename is somehow used for looking up fragments within the resolver, at which time a fragment on a concrete type (which is totally valid) would still need typename via the fragment. Leaving the status quo seemed safest.

plan.go
gmac: I actually propose pulling that method entirely in https://github.com/movio/bramble/pull/94. `selectionSetHasFieldNamed ` is faulty as written because it doesn’t take field aliases into account. If there’s human sensitivity to field duplication in the selection (machines don’t care), I’d propose just running a loop once at the end of extractSelections that does a single-pass uniqueness on alias falling back to name.

nmaquet: Ok, that's fair. Let's worry about that at a later stage then (if needed).

nmaquet (approved):
Thanks @gmac!

Issue #90, opened 2021-11-04
Proposed fix: lock down aliases that may break queries


gmac:
## The Problem

Field aliases are extremely difficult in federation design because they permit clients to break their own requests by manipulating reserved selection hints. Case of point, here's a busted query:

```graphql
query {
  shop1 {
    products {
      boo: id
      name
    }
  }
}
```

Here's another...

```graphql
query {
  shop1 {
    products {
      id: name
    }
  }
}
```

And another...

```graphql
query {
  shop1 {
    products {
      id
      _id: name
    }
  }
}
```

As demonstrated, there are lots of creative ways to disrupt key selections. This same situation applies when `__typename` is hijacked as a custom alias on abstract types.

## Proposed solution

To lock this down, validation errors are necessary to reserve key aliases for implementation details. I would propose that  `__id` and `__typename` are standardized and reserved as system aliases on boundary types (current `_id` becomes `__id`, and the double-underscore becomes a convention of reserved system fields).

With that, errors are inserted into the query planner to invalidate requests that conflict with reserved aliases:

```go
// while planning field selections...
if selection.Alias == "__id" && selection.Name != "id" {
  gqlerror.Errorf("invalid alias: \"__id\" is a reserved alias on type %s", parentType)
}
if selection.Alias == "__typename" && selection.Name != "__typename" {
  gqlerror.Errorf("invalid alias: \"__typename\" is a reserved alias on type %s", parentType)
}
```

Lastly, `__id` is _always_ added to boundary types, and `__typename` is _always_ added to abstract types (there's [a PR in the queue](https://github.com/gmac/bramble/pull/1) for the later change).

@nmaquet and @lucianjon – feelings on this proposal? If we're aligned on the approach, I'll put a PR together after https://github.com/movio/bramble/pull/89 goes out.

yaacovCR:
>All fields defined within an Object type must not have a name which begins with "__" (two underscores), as this is used exclusively by GraphQL’s introspection system.

https://spec.graphql.org/draft/#sec-Objects

gmac:
Hmm. Fair. @yaacovCR how does Tools stitching handle this situation of user-defined selections conflicting with selection sets. IE: if there’s a user selection of “id: name”, and a selectionSet of “{ id }”, then how is that reconciled?

nmaquet:
[...]

PR #92, opened 2021-11-05
Fix: validate reserved field aliases

gmac:
Imposes validation errors around the use of `__id` and `__typename` aliases to prevent user-sabotaged query errors, as discussed in https://github.com/movio/bramble/issues/90.

## The Problem

Per https://github.com/movio/bramble/issues/90, there were lots of ways to hoodwink `id` and `__typename` selection hints into breaking a query, for example:

```graphql
query {
  shop1 {
    products {
      boo: id
      name
    }
  }
}
```

The permissiveness of GraphQL aliases allows a user to hijack federation implementation details.

## The Fix

This adds validation errors to restrict the use of `__id` and `__typename` field aliases for non-standard purposes. In this process, `_id` becomes `__id` to align with the double-underscore convention of GraphQL internals.

Resolves https://github.com/movio/bramble/issues/90

### DIFF
--- a/docs/algorithms.md
+++ b/docs/algorithms.md
@@ -260,8 +260,8 @@ function ExtractSelectionSet(ctx, insertionPoint, parentType, selectionSet, loca
             append childrenSteps to the result's query plan steps
         }
     }
-    if parentType is a boundary type and the result selectionSet doesn't have an "id" field {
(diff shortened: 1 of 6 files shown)


gmac:
Closing this one in favor of a more comprehensive fix to the problem.

Issue #93, opened 2021-11-05
Broken boundary when aliasing id field


gmac:
## The Problem

Similar to https://github.com/movio/bramble/issues/90. There still appears to be sensitivity around `id` being aliased within the execution pipeline, even while running the reserved namespace safeguards in https://github.com/movio/bramble/pull/92. For example, this query still fails:

```graphql
query {
  shop1 {
    products {
      id: name
    }
  }
}
```

Here's the message:

```json
{
  "message": "got a null response for non-nullable field \"id\"",
  "path": [
    "shop1",
    "products",
    0,
    "id"
  ]
}
```

This issue is very specific to the `id` field, which suggests that [the special execution handling](https://github.com/movio/bramble/blob/main/query_execution.go#L523) of the `id` field causes the problem. Ideally, `id` should operate as a normal member of the GraphQL schema without restrictions.

## Proposed fix

The work in https://github.com/movio/bramble/pull/92 is a big step in the right direction here, however it leaves the [special execution rule](https://github.com/movio/bramble/blob/main/query_execution.go#L523) around `id` in place. To close this out once and for all, I'd suggest removing _all_ special considerations from the `id` field and to rely exclusively on the reserved `__id` namespace for federation mapping. With that, the complete schema will be an open book for standard GraphQL patterns, and only `__id` and `__typename` will impose restrictions about customizing their use.

nmaquet:
Great spotting @gmac. I'll have a chat with @lucianjon on what would be the best way to address this.

gmac:
The implementation of #90 in https://github.com/movio/bramble/pull/94 also fixes this. I’ll tidy up that PR later tonight and formally open it for consideration.

PR #94, opened 2021-11-06
Fix: user-defined field aliases may break federation

gmac:
This adds `_bramble_id` as a universal ID hint, which allows `id` to behave as a standard user-defined selection. This fixes the numerous ways to break federation with user-defined field aliases describe in https://github.com/movio/bramble/issues/90 and https://github.com/movio/bramble/issues/93.

Apologies for the large PR – the vast majority of it is test fixtures.

## The Problem

As described in https://github.com/movio/bramble/issues/90, there are many creative ways for user-defined queries to break Bramble federation ids using field aliases, for example:

```graphql
query {
  shop1 {
    products {
      boo: id
      name
    }
  }
}
```

While dynamic adjustments _can_ be made to the query planner to account for such aliases, a user-aliased id still ends up breaking in the execution resolver, per https://github.com/movio/bramble/issues/93.

A general observation: query planning currently optimizes for vanity queries at the cost of tricky logic holes and extra looping. The process becomes simpler and more reliable when the query planner always adds consistent implementation fields for itself, even if they are redundant with the contents of the user-defined query. Not making `id` do double-duty as both a user selection and an implementation detail avoids contention between the planning and resolution steps.

## The fix(es)

- Makes the planner _always_ add a `_bramble_id: id` selection to boundary scopes. This becomes a universal implementation key, eliminating `_id` and leaving `id` untouched as a basic user selection.
- Validates `_bramble_id` and `__typename` aliases. A user selection will not be allowed to hijack these fields.
- Foregoes vanity queries in favor of eliminating `childstep` recursion complexity and `selectionSetHasFieldNamed` looping. Queries are now more verbose, but they are extremely consistent and avoid [suspected logic holes](https://github.com/movio/bramble/pull/94#discussion_r744355229).

## Duplicated selections are ugly

If there’s human sensitivity to duplicated field selections (machines don’t care), then I’d propose adding a single-pass loop at the end of extractSelectionSet that imposes uniqueness _once_ based on field alias falling back to field name. At present, “selectionSetHasFieldNamed” runs for each de-duplicated field, and doesn’t take into account field aliases.

Resolves https://github.com/movio/bramble/issues/90.
Resolves https://github.com/movio/bramble/issues/93.

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/vektah/gqlparser/v2/ast"
+	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
 // QueryPlanStep is a single execution step
@@ -66,7 +67,7 @@ func Plan(ctx *PlanningContext) (*QueryPlan, error) {
 		return nil, fmt.Errorf("not implemented")
 	}
 
-	steps, err := createSteps(ctx, nil, parentType, "", ctx.Operation.SelectionSet, false)
+	steps, err := createSteps(ctx, nil, parentType, "", ctx.Operation.SelectionSet)
 	if err != nil {
 		return nil, err
 	}
@@ -75,7 +76,7 @@ func Plan(ctx *PlanningContext) (*QueryPlan, error) {
 	}, nil
 }
 
-func createSteps(ctx *PlanningContext, insertionPoint []string, parentType, parentLocation string, selectionSet ast.SelectionSet, childstep bool) ([]*QueryPlanStep, error) {
+func createSteps(ctx *PlanningContext, insertionPoint []string, parentType string, parentLocation string, selectionSet ast.SelectionSet) ([]*QueryPlanStep, error) {
 	var result []*QueryPlanStep
 
 	routedSelectionSet, err := routeSelectionSet(ctx, parentType, parentLocation, selectionSet)
@@ -84,7 +85,7 @@ func createSteps(ctx *PlanningContext, insertionPoint []string, parentType, pare
 	}
 
 	for location, selectionSet := range routedSelectionSet {
-		selectionSetForLocation, childrenSteps, err := extractSelectionSet(ctx, insertionPoint, parentType, selectionSet, location, childstep)
+		selectionSetForLocation, childrenSteps, err := extractSelectionSet(ctx, insertionPoint, parentType, selectionSet, location)
 
 		if err != nil {
 			return nil, err
@@ -115,13 +116,23 @@ func createSteps(ctx *PlanningContext, insertionPoint []string, parentType, pare
 	return result, nil
 }
 
-func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentType string, input ast.SelectionSet, location string, childstep bool) (ast.SelectionSet, []*QueryPlanStep, error) {
+var reservedAliases = map[string]string{
+	"__typename": "__typename",
+	"_bramble_id": "id",
+}
+
+func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentType string, input ast.SelectionSet, location string) (ast.SelectionSet, []*QueryPlanStep, error) {
 	var selectionSetResult []ast.Selection
 	var childrenStepsResult []*QueryPlanStep
 	var remoteSelections []ast.Selection
 	for _, selection := range input {
 		switch selection := selection.(type) {
 		case *ast.Field:
+			for reservedAlias, requiredName := range reservedAliases {
+				if selection.Alias == reservedAlias && selection.Name != requiredName {
+					return nil, nil, gqlerror.Errorf("%s.%s: alias \"%s\" is reserved for system use", strings.Join(insertionPoint, "."), reservedAlias, reservedAlias)
+				}
+			}
 			if parentType != queryObjectName && parentType != mutationObjectName && ctx.IsBoundary[parentType] && selection.Name == "id" {
 				selectionSetResult = append(selectionSetResult, selection)
 				continue
@@ -129,7 +140,7 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
 			loc, err := ctx.Locations.URLFor(parentType, location, selection.Name)
 			if err != nil {
 				// namespace
-				subSS, steps, err := extractSelectionSet(ctx, append(insertionPoint, selection.Name), selection.Definition.Type.Name(), selection.SelectionSet, location, childstep)
+				subSS, steps, err := extractSelectionSet(ctx, append(insertionPoint, selection.Name), selection.Definition.Type.Name(), selection.SelectionSet, location)
 				if err != nil {
 					return nil, nil, err
 				}
@@ -153,7 +164,6 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
 					selection.Definition.Type.Name(),
 					selection.SelectionSet,
 					location,
-					childstep,
 				)
 				if err != nil {
 					return nil, nil, err
@@ -169,14 +179,11 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
 				selection.TypeCondition,
 				selection.SelectionSet,
 				location,
-				childstep,
 			)
 			if err != nil {
 				return nil, nil, err
 			}
-			if !selectionSetHasFieldNamed(selectionSet, "__typename") {
-				selectionSet = append(selectionSet, &ast.Field{Alias: "__typename", Name: "__typename", Definition: &ast.FieldDefinition{Name: "__typename", Type: ast.NamedType("String", nil)}})
-			}
+			selectionSet = append(selectionSet, &ast.Field{Alias: "__typename", Name: "__typename", Definition: &ast.FieldDefinition{Name: "__typename", Type: ast.NamedType("String", nil)}})
 			inlineFragment := *selection
 			inlineFragment.SelectionSet = selectionSet
 			selectionSetResult = append(selectionSetResult, &inlineFragment)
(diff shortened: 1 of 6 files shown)


plan_test.go
```diff
@@ -844,11 +818,19 @@ func TestQueryPlanNoUnnessecaryID(t *testing.T) {
 		  {
 			"ServiceURL": "A",
 			"ParentType": "Query",
-			"SelectionSet": "{ movies { title } }",
+			"SelectionSet": "{ movies { title _bramble_id: id } }",
```
gmac, line 821: I believe this test here represented another broken scenario: the boundary type (Movie) was being requested without an `id`. Digging into the planning flow, the omission was caused by a nuance in the `childstep` tracking. I've eliminated the `childstep` logic entirely as its sole purpose appears to have been controlling selection redundancies.

plan.go
```diff
@@ -66,7 +67,7 @@ func Plan(ctx *PlanningContext) (*QueryPlan, error) {
 		return nil, fmt.Errorf("not implemented")
 	}
 
-	steps, err := createSteps(ctx, nil, parentType, "", ctx.Operation.SelectionSet, false)
+	steps, err := createSteps(ctx, nil, parentType, "", ctx.Operation.SelectionSet)
```
gmac, line 70: I've eliminated the `childstep` recursion flag because its only purpose was to conditionally manage the inclusion of federation `id`s – to which I'm [not convinced it wasn't buggy](https://github.com/movio/bramble/pull/94/files#r744355229). Now that `_bramble_id` is ubiquitously included in all boundary scopes, the complexity of `childstep` is unnecessary.

plan.go
```diff
@@ -115,21 +116,32 @@ func createSteps(ctx *PlanningContext, insertionPoint []string, parentType, pare
...
 	var childrenStepsResult []*QueryPlanStep
 	var remoteSelections []ast.Selection
 	for _, selection := range input {
 		switch selection := selection.(type) {
 		case *ast.Field:
+			for reservedAlias, requiredName := range reservedAliases {
+				if selection.Alias == reservedAlias && selection.Name != requiredName {
+					return nil, nil, gqlerror.Errorf("%s.%s: alias \"%s\" is reserved for system use", strings.Join(insertionPoint, "."), reservedAlias, reservedAlias)
+				}
+			}
```
gmac, line 135: This loop and the `reservedAliases` struct above block the submission of reserved aliases. If the user attempts to make a selection that will interfere with internal Bramble operations, we'll simply tell them it's not allowed.

plan.go
```diff
@@ -169,14 +180,11 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
...
 				location,
-				childstep,
 			)
 			if err != nil {
 				return nil, nil, err
 			}
-			if !selectionSetHasFieldNamed(selectionSet, "__typename") {
-				selectionSet = append(selectionSet, &ast.Field{Alias: "__typename", Name: "__typename", Definition: &ast.FieldDefinition{Name: "__typename", Type: ast.NamedType("String", nil)}})
-			}
+			selectionSet = append(selectionSet, &ast.Field{Alias: "__typename", Name: "__typename", Definition: &ast.FieldDefinition{Name: "__typename", Type: ast.NamedType("String", nil)}})
```
gmac, line 186: The `selectionSetHasFieldNamed` function was buggy because it didn't account for field aliases. Rather than expanding its implementation to handle more complex naming considerations, I've opted to remove it entirely given that it was adding loops mostly for the sake of vanity selections. A `__typename` field will now _always_ be appended to fragments, even if it was also included in the user-defined selection.

plan.go
```diff
@@ -247,40 +252,28 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
...
-	} else if parentType != queryObjectName &&
-		parentType != mutationObjectName &&
-		ctx.IsBoundary[parentType] &&
-		(childstep || len(childrenStepsResult) > 0) &&
-		parentDef.Fields.ForName("id") != nil &&
-		!selectionSetHasFieldNamed(selectionSetResult, "id") {
-		id := &ast.Field{
-			Alias:      "_id",
-			Name:       "id",
-			Definition: parentDef.Fields.ForName("id"),
```
gmac, line 276: This check has been greatly simplified – any boundary object with an `id` field defined gets a `_bramble_id` hint added to the end of the selection. It's more verbose, but it's extremely predictable and consistent.

plan.go
```diff
@@ -247,40 +252,28 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
...
-					TypeCondition: implementationName,
-					SelectionSet: []ast.Selection{
-						&ast.Field{
-							Alias:      "_id",
-							Name:       "id",
-							Definition: implementationType.Fields.ForName("id"),
-						},
-					},
-					ObjectDefinition: implementationType,
+
```
gmac, line 254: This adds the best-practice from [below](https://github.com/movio/bramble/pull/94/files#r744360598) where the type is checked for an ID definition – I suppose the wild cards here are namespaces, which are just merged scopes without ids?

plan.go
```diff
@@ -115,21 +116,31 @@ func createSteps(ctx *PlanningContext, insertionPoint []string, parentType, pare
 	return result, nil
 }
 
-func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentType string, input ast.SelectionSet, location string, childstep bool) (ast.SelectionSet, []*QueryPlanStep, error) {
+var reservedAliases = map[string]string{
+	"__typename": "__typename",
+	"_bramble_id": "id",
+}
```
gmac, line 122: This is a mapping of reserved aliases – or, alias/name pairs that are sensitive to the implementation, and therefore will be validated on the gateway for integrity.

plan.go
```diff
@@ -247,40 +251,28 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
...
-					},
-					ObjectDefinition: implementationType,
+
+				if idDef := implementationType.Fields.ForName("id"); idDef != nil {
+					possibleId := &ast.InlineFragment{
+						TypeCondition: implementationName,
+						SelectionSet: []ast.Selection{&ast.Field{Alias: "_bramble_id", Name: "id", Definition: idDef}},
+						ObjectDefinition: implementationType,
+					}
+					selectionSetResult = append(selectionSetResult, possibleId)
```
gmac, line 261: This adds the best-practice from [below](https://github.com/movio/bramble/pull/94/files#r744360598) where the type is checked for an ID definition – I suppose the wild cards here are namespaces, which are just merged scopes without ids?

plan_fixtures_test.go
```diff
@@ -264,7 +264,7 @@ var PlanTestFixture6 = &PlanTestFixture{
 }
 
 
-func (f *PlanTestFixture) Plan(t *testing.T, query string) *QueryPlan {
+func (f *PlanTestFixture) Plan(t *testing.T, query string) (*QueryPlan, error) {
```
gmac, line 267: I've broken these test operations a bit more to add an additional check for error outcomes.

plan_test.go
```diff
@@ -357,32 +357,6 @@ func TestQueryPlanFragmentSpread1(t *testing.T) {
 	PlanTestFixture1.Check(t, query, plan)
 }
 
-func TestQueryPlanFragmentSpread1DontDuplicateTypename(t *testing.T) {
```
gmac, line 360: This test is no longer applicable now that `__typename` is, by design, always appended and allowed to repeat a user-defined selection.

query_execution.go
```diff
@@ -520,7 +520,7 @@ func mergeExecutionResultsRec(src interface{}, dst interface{}, insertionPoint [
 					}
 					if srcID == dstID {
 						for k, v := range result {
-							if k == "_id" || k == "id" {
+							if k == "_bramble_id" {
```
gmac, line 523: Then here's where https://github.com/movio/bramble/issues/93 was breaking... this special case for the `id` field prevented it from resolving standard GraphQL aliases. Now that everything revolves around `_bramble_id`, we can leave `id` alone an allow it to resolve as a normal field.

gmac:
Okay, ready for review here @nmaquet. Sorry for the large PR.

lucianjon:
@gmac Thanks again for this one, have just had a skim so far but looks solid. One thought, if we're going the route of reserving the `__typename` and `_bramble_id` aliases, as well as guaranteeing the injection of `__typename`. What do you think about following the same pattern as the id field and making it something like `_bramble__typename`, this could also allow us to drop it from the final response like we do the id.

It would require updating the execution pipeline to look for that alias over the regular `__typename` but would have the benefit of clearly indicating it's been injected for the internal purposes of bramble.

gmac:
@lucianjon — a dedicated typename alias makes sense to me. It’d certainly be ideal to keep __typename itself as native GraphQL to be abused in any way the end user likes.

plan.go
suessflorian: Hey I was curious about this part here so did some testing - I was mainly wondering if this would effect `__typename` fields in selection sets. Seems to be fine, moreover, it's fine with `__typename: __typename` which I thought was fishy. Then I tried out `_bramble_id: id`, this also works without error. Perhaps this isn't working as intended?

Here's a test (for `plan_test.go`):
```go
func TestQueryBrambleReserved(t *testing.T) {
	query := `{
		movies {
			... on Movie {
				id
				title(language: French)
				_bramble_id: id
			}
		}
	}`
	plan := `{
		"RootSteps": [
			{
				"ServiceURL": "A",
				"ParentType": "Query",
        "SelectionSet": "{ movies { ... on Movie { id title(language: French) _bramble_id: id } _bramble_id: id } }",
				"InsertionPoint": null,
				"Then": null
			}
		]
	}`
	PlanTestFixture1.Check(t, query, plan) // we actually expect a panic here, but looking at the diff here is useful
}
```

suessflorian: Ah nevermind this, didn't quite see the last part of the condition being a `!=`. Sorry for the noise.

lucianjon:
> @lucianjon — a dedicated typename alias makes sense to me. It’d certainly be ideal to keep __typename itself as native GraphQL to be abused in any way the end user likes.

Sweet, I think we could do it in another PR to keep the size of this one down.

plan.go
```diff
@@ -247,40 +251,28 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
...
-							Name:       "id",
-							Definition: implementationType.Fields.ForName("id"),
-						},
-					},
-					ObjectDefinition: implementationType,
+
+				if idDef := implementationType.Fields.ForName("id"); idDef != nil {
+					possibleId := &ast.InlineFragment{
+						TypeCondition: implementationName,
+						SelectionSet: []ast.Selection{&ast.Field{Alias: "_bramble_id", Name: "id", Definition: idDef}},
```
lucianjon, line 258: Could pay to pull `_bramble_id` out into a const since it's used in a few spots.

lucianjon (approved):
Looks good from my perspective, will give someone else the chance to have a look though.

nmaquet:
Thanks @gmac!!!

Issue #87, opened 2021-11-03
Repeated selections for distinct field names are dropped during merge


gmac:
Low-priority edge case: when making a query selection with multiple selections on a distinct field, data is only returned for the first selection...

```graphql
query {
  shop1 {
    products {
      name
    }
    products {
      manufacturer {
        name
      }
    }
  }
}
```

This query only returns data for the first field selection, while technically it should return a union of the two:

```json
{
  "data": {
    "shop1": {
      "products": [
        {
          "name": "iPhone"
        },
        {
          "name": "Apple Watch"
        }
      ]
    }
  }
}
```

The query plan looks about like I'd expect, and my remote servers are executing the appropriate steps:

```json
{
    "RootSteps": [{
        "ServiceURL": "http://localhost:4003/graphql",
        "ParentType": "Query",
        "SelectionSet": "{ shop1 { _id: id products { _id: id } products { _id: id } } }",
        "InsertionPoint": null,
        "Then": [{
            "ServiceURL": "http://localhost:4002/graphql",
            "ParentType": "Product",
            "SelectionSet": "{ _id: id name }",
            "InsertionPoint": ["shop1", "products"],
            "Then": null
        }, {
            "ServiceURL": "http://localhost:4002/graphql",
            "ParentType": "Product",
            "SelectionSet": "{ _id: id manufacturer { _id: id } }",
            "InsertionPoint": ["shop1", "products"],
            "Then": [{
                "ServiceURL": "http://localhost:4001/graphql",
                "ParentType": "Manufacturer",
                "SelectionSet": "{ _id: id name }",
                "InsertionPoint": ["shop1", "products", "manufacturer"],
                "Then": null
            }]
        }]
    }]
}
```

Therefore, I'd venture to guess that the loss of data happens in the post-request assembly process. Logging this here as a backlog item.

nmaquet:
Thanks for reporting this! We'll look into it in the coming days.

gmac:
Suspicions point to `formatResponseData` being the source of error. Looks like merged data makes it through with the expected map keys, so the error must happen while building the final data structure based from the selection set.

PR #104, opened 2021-11-09
Fix: merge composite scopes in final selection

gmac:
Fixes a bug with the final filtering of non-distinct composite field selections, per https://github.com/movio/bramble/issues/87.

## The Problem

Multiple selection scopes applied to the same object selection were not getting combined in the final result, for example:

```graphql
query {
  shop1 {
    products {
      id
    }
    products {
      name
    }
  }
}
```

While this query properly _fetched_ all requested data from underlying services (i.e.: query planning worked as expected), the resulting data only included fields for the first distinct `products` selection scope:

```json
{
  "data": {
    "shop1": {
      "products": [
        {
          "id": "gid://shopapp/Product/7"
          // ack, no "name"
        }
      ]
    }
  }
}
```

## The Fix

This adjusts `unionAndTrimSelectionSetRec`, where the mapping of "seenFields" was filtering all repeated selection names as though they were leaf values. This new implementation has repeat composite fields merge their selections together.

Resolves https://github.com/movio/bramble/issues/87.

### DIFF
--- a/execution_test.go
+++ b/execution_test.go
@@ -2233,6 +2233,55 @@ func TestUnionAndTrimSelectionSet(t *testing.T) {
 		require.Equal(t, selectionSet, filtered)
 	})
 
+	t.Run("removes duplicate leaf values and merges composite scopes", func(t *testing.T) {
+		selectionSet := ast.SelectionSet{
+			&ast.Field{
+				Alias:            "name",
+				Name:             "name",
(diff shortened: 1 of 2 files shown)


pkqk (approved):
Good find

lucianjon:
Wicked, thanks @gmac

PR #107, opened 2021-11-12
Refactor: interface selections have proper location

gmac:
As described in https://github.com/movio/bramble/issues/113, it's less than ideal that namespaces and interfaces are omitted from the `Locations` map, and therefore get [handled as error side-effects](https://github.com/movio/bramble/blob/main/plan.go#L140-L141):

```go
loc, err := ctx.Locations.URLFor(parentType, location, selection.Name)
// Errors are returned for unmapped namespace/interface locations
if err == nil && loc != location {
  // ...
}
```

These missing locations bank on a not-obvious rule that they're not allowed to transition services, so can alway use their parent object's location. Ideally these known resources would simply route to known locations, and missing location errors would be treated as _actual_ errors.

This adds interfaces to the `Locations` map so they become a properly routed resource.

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -59,7 +59,7 @@ func buildFieldURLMap(services ...*Service) FieldURLMap {
 	result := FieldURLMap{}
 	for _, rs := range services {
 		for _, t := range rs.Schema.Types {
-			if t.Kind != ast.Object || isGraphQLBuiltinName(t.Name) || t.Name == serviceObjectName {
+			if (t.Kind != ast.Object && t.Kind != ast.Interface) || isGraphQLBuiltinName(t.Name) || t.Name == serviceObjectName {
 				continue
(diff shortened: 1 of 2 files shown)


lucianjon:
Sorry this one took a while to get to @gmac, will take a look over the next day or so if someone doesn't get to it first

lucianjon:
Thanks @gmac, even if it's optional at this point I think this is still definitely worthwhile. I actually have a hunch this will work towards fixing a bug with some internal tooling that relies on the [meta plugin](https://github.com/movio/bramble/blob/main/plugins/meta.go#L247). I'll have a play with it after merging this in but I assume I'll need to add `interface` to the case statement [here](https://github.com/movio/bramble/blob/main/plugins/meta.go#L185) too.

Issue #106, opened 2021-11-10
Filtering base interface selections is not type-aware


gmac:
Another crazy one along the lines of https://github.com/movio/bramble/issues/87, and confirms the issue hypothesized in https://github.com/movio/bramble/pull/91.

## The Problem

During final selection filtering in the execution pipeline, type awareness seems to be lost on base interface selections (i.e.: fields selected from an interface without going through a typed fragment). For example, here `myInterface` is an interface type that offers `myUnion` as a base field:

```graphql
query {
  myInterface {
    myUnion {
      ...on Product {
        name
        manufacturer {
          name
        }
      }
    }
  }
}
```

Stacking these abstract selections atop one another results in the following unsuccessful outcome... while the underlying queries look accurate, the final filtered result is off:

```json
{
  "data": {
    "myInterface": {
      "myUnion": {
        "name": "iPhone",
        "manufacturer": {
          "_bramble_id": "1"
        },
        "_bramble_id": "1",
        "_bramble__typename": "Product"
      }
    }
  }
}
```

HOWEVER, this issue can be avoided by wrapping the second-level union in a typed fragment. The following selection works perfectly, implying that the problem is specifically with base interface selections:

```graphql
query {
  myInterface {
    ...on Product {
      myUnion {
        ...on Product {
          name
          manufacturer {
            name
          }
        }
      }
    }
  }
}
```

## Probable cause

I believe this requires a complimenting fix to the changes made in https://github.com/movio/bramble/pull/91. While https://github.com/movio/bramble/pull/91 assured that abstract types always have access to their type information, I believe that the execution pipeline must still be updated to leverage this base type information.

gmac:
Update...

## Actual cause

The `myInterface` field was being treated as a namespace, and the namespace handler block was modifying the original user selection during planning rather than adjusting a copy. This is addressed in https://github.com/movio/bramble/pull/108.

The fact that `myInterface` was being treated as a namespace is debatably problematic, though not strictly broken. A proposed fix for supporting interfaces with first-class routing is in https://github.com/movio/bramble/pull/107.

PR #108, opened 2021-11-12
Fix: broken namespace aliases and mutable selections

gmac:
Fixes some problems with namespace operations in the query planner, and simplifies the overall logic in the process.

## Problems + Fixes

```graphql
query {
  boo: myNamespace {
    product {
      name
    }
    manufacturer {
      name
    }
  }
}
```

* The namespace handler condition was demonstrably broken by field aliases: it [composed its insertion point](https://github.com/movio/bramble/pull/108/files#diff-11ffedf6a81044546aa19c8009998208f5a7afec3cdff04ca4281117b5ec0da7L143) using field `Name` rather than `Alias`. This issue is now fixed.

* The namespace condition could also corrupt the user's original selection because it modified fields directly rather than making copies. This was problematic because it modified the selection used to filter the final result, as described in https://github.com/movio/bramble/issues/106. This is also fixed.

* With the above adjustments made, the namespace condition ended up basically identical to the normal composite field handler. So, this simplifies the logic so that a failed location lookup now just flows into the normal field handlers; this new pattern has the added advantage of handing both leaf values and composite fields.

Resolves https://github.com/movio/bramble/issues/106 by happy accident.

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -138,18 +138,8 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
 				continue
 			}
 			loc, err := ctx.Locations.URLFor(parentType, location, selection.Name)
-			if err != nil {
-				// namespace
-				subSS, steps, err := extractSelectionSet(ctx, append(insertionPoint, selection.Name), selection.Definition.Type.Name(), selection.SelectionSet, location)
-				if err != nil {
-					return nil, nil, err
-				}
-				selection.SelectionSet = subSS
-				selectionSetResult = append(selectionSetResult, selection)
-				childrenStepsResult = append(childrenStepsResult, steps...)
-				continue
-			}
-			if loc != location {
+			// Errors are returned for unmapped namespace/interface locations (needs refactor)
+			if err == nil && loc != location {
 				// field transitions to another service location
 				remoteSelections = append(remoteSelections, selection)
 			} else if selection.SelectionSet == nil {
@@ -157,7 +147,6 @@ func extractSelectionSet(ctx *PlanningContext, insertionPoint []string, parentTy
 				selectionSetResult = append(selectionSetResult, selection)
 			} else {
 				// field is a composite type in the current service
-				newField := *selection
 				selectionSet, childrenSteps, err := extractSelectionSet(
 					ctx,
 					append(insertionPoint, selection.Alias),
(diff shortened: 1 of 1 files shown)


lucianjon:
Thanks @gmac, looks like an improvement on what we had. I'd like to refactor that whole piece at some point, as assuming the error is actually not an error makes the conditions read quite weird. That would require another way to figure out the field requested is a namespace, which I'm sure is possible but just requires some thinking. 

For now would you mind explicitly checking for `err == nil` (a namespace will never be a leaf) in the `else if` case. As well as chucking a comment in the else case to say the error is ignored as it's assumed to be a namespace if `err != nil`. Just to make sure someone else coming along doesn't think the error is unintentionally ignored. We're still ignoring some potential actual errors but I think it's the best that can be done for now and the query validation should be catching the bulk of those.

gmac:
@lucianjon – comment added! I had the same thought about that error line, the logic is weird and goes against the grain of exceptions being exceptional. For a refactor, I've sort of been thinking that the `isBoundary` map could become more robust with other useful type information, for example:

```json
FederationObjects: {
  MyNamespace: { Kind: "namespace", Implementations: [] },
  MyBoundary: { Kind: "boundary", Implementations: [] },
  MyInterface: { Kind: "interface", Implementations: ["MyBoundary"] },
}
```

That kind of structure would be able to distinguish boundaries and namespaces, and would provide a pre-compiled cache of all abstract boundaries so that they [wouldn't have to be enumerated repeatedly](https://github.com/movio/bramble/blob/main/plan.go#L245-L252).

lucianjon:
> @lucianjon – comment added! I had the same thought about that error line, the logic is weird and goes against the grain of exceptions being exceptional. For a refactor, I've sort of been thinking that the `isBoundary` map could become more robust with other useful type information, for example:
> 
> ```go
> FederationObjects: {
>   MyNamespace: { Kind: "namespace", Implementations: nil },
>   MyBoundary: { Kind: "boundary", Implementations: nil },
>   MyInterface: { Kind: "interface", Implementations: ["MyBoundary"] },
> }
> ```
> 
> That kind of structure would be able to distinguish boundaries and namespaces, and would provide a pre-compiled cache of all abstract boundaries so that they [wouldn't have to be enumerated repeatedly](https://github.com/movio/bramble/blob/main/plan.go#L245-L252).

Wicked cheers, will get this in. Yeah agreed on the refactor, we have access to all the information we need when we merge schemas and could easily just keep it all stored on there. Can write an issue up for it or if you'd like to, go for it!

gmac:
@lucianjon – https://github.com/movio/bramble/issues/113

PR #114, opened 2021-11-17
Add handling of ast.Interface to kindToStr

lucianjon:
The previous case statement wasn't actually exhaustive as it excluded `interface`.  Now if `kindToStr` gets an interface it'll return the correct string. The special case for `ast.InputObject` and `ast.Object` were kept for backwards compatibility.


gmac (approved):
Tidy.

Issue #109, opened 2021-11-12
Support configuration of the "id" key name


gmac:
The fact that Bramble hard-codes the federation key name as `id` is prohibitively strict; and it should be pretty simple to make that configurable. This would unlock Bramble for graphs that don't call their primary key `id`.

Interesting backstory – I actually worked on such a graph. In its early days as a collection of decoupled services, each type in each service had implemented its own unique `id` signature – sometimes as `Int`, sometimes as `String`, sometimes as `ID`. Once our federation intentions were solidified, it was already too late to standardize on one `id` field format because so much was built around the existing type-specific IDs. Thus, a new field called `uid` (unique id) was introduced across the graph as the federation key. It would be a shame not to support such a graph simply due to its history around the name "id".

nmaquet:
@gmac What you're suggesting would be to make the federation key name configurable globally for the entire data graph, correct? Or were you thinking about a key name per object type?

gmac:
While customizable keys per type are _useful_, I don't think they're strictly necessary here... Bramble isn't trying to win the customizability battle against the likes of Apollo Federation and Schema Stitching. However, it'd be a real bummer to have a graph of mutually-exclusive schemas keyed with a common field (a perfect fit for Bramble!) to be blocked simply because the common key name was something other than the string "id". Making it configurable once for the graph should be pretty simple.

Issue #97, opened 2021-11-07
RFC: require all boundary accessor fields to return a nullable result.


nmaquet:
Supporting non-nullable boundary accessors adds a significant amount of complexity to the implementation of Bramble's query executor. This is due to [GraphQL's bubbling of errors to parents of non-nullable fields](https://spec.graphql.org/June2018/#sec-Errorsl). Dropping support for nullable boundary accessor fields would make Bramble quite a bit simpler.

nmaquet:
@gmac Curious to hear your thoughts on this one.

gmac:
Hmmm. What are you referring to as a boundary accessor? A root accessor specifically used for federation access, or any object field in the graph that relates to a boundary type? Taking this in turn:

- Root accessors used for federation should absolutely be nullable; single-records should always use “thing(id: ID!): Thing”, and lists should always use “things(ids: [ID!]!): [Thing]!” (not-null list with nullable entries). Not using those signatures causes avoidable problems, so requiring them nudges developers into the pit of success, so win!
- However, forcing any object association in the graph that mentions a boundary type to be not-null feels like a no-go to me. While I whole-heartedly agree that associations that cross boundaries _should_ be nullable to avoid avoidable null bubbling, politics sometimes win out on the call to make an association not-null. Also there’s the situation of boundary types referencing each other within the same local graph where they’re unaware of reasons why they _shoudn’t_ be allowed to hard-reference. In general, this situation feels like a limitation that works against general GraphQL.

SO, if we’re just talking root accessors, this sounds totally reasonable. If it’s constraining standard GraphQL patterns across the graph though, that’d probably be grounds for forking by third-parties.

nmaquet:
@gmac Thanks for your answer. I was definitely meaning root accessor fields used for federation access.
>  Not using those signatures causes avoidable problems, so requiring them nudges developers into the pit of success, so win!
100% agreed!

gmac:
One other though here related to https://github.com/movio/bramble/issues/109: it'd be a lot more flexible for integrators if the accessor argument name was read from the schema the same way the field name is. Then boundary accessors would be mapped as `{ Field: "myAccessor", Argument: "id", Array: false }`.

nmaquet:
> One other though here related to #109: it'd be a lot more flexible for integrators if the accessor argument name was read from the schema the same way the field name is. Then boundary accessors would be mapped as `{ Field: "myAccessor", Argument: "id", Array: false }`.

[...]

PR #115, opened 2021-11-17
Dynamic boundary argument

gmac:
As discussed in https://github.com/movio/bramble/issues/97#issuecomment-968905753, this reads the name of each boundary query argument from the schema rather than hard-coding the argument names as `id` and `ids`. Now a boundary query naming is freeform so long as the field conforms to the basic signature of input and return:

```graphql
getMyThings(thingIds: [ID!]!): [Thing]! @boundary
```

This is an important first step in allowing the federation key to be customized for the graph, per https://github.com/movio/bramble/issues/109.


lucianjon:
@gmac No problems here, lets get it in

PR #116, opened 2021-11-20
Allow customization of the graph "id" field

gmac:
Per https://github.com/movio/bramble/issues/109, it was a bit restrictive that Bramble mandated the name `id` as the only allowed graph key. This adds configuration to customize the package's `IdFieldName`, at which time a graph may specify its own universal boundary key name. Also updates related docs.

Resolves https://github.com/movio/bramble/issues/109.

### DIFF
--- a/config.go
+++ b/config.go
@@ -22,6 +22,7 @@ type PluginConfig struct {
 
 // Config contains the gateway configuration
 type Config struct {
+	IdFieldName            string    `json:"id-field-name"`
 	GatewayListenAddress   string    `json:"gateway-address"`
 	MetricsListenAddress   string    `json:"metrics-address"`
 	PrivateListenAddress   string    `json:"private-address"`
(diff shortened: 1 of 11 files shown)


config.go
```diff
@@ -93,6 +94,10 @@ func (c *Config) Load() error {
 	}
 	c.Plugins = plugins
 
+	if c.IdFieldName != "" {
```
pkqk, line 97:
```suggestion
	if strings.TrimSpace(c.IdFieldName) != "" {
```
it's possibly worth being defensive here in case some empty but non blank strings sneak their way into the config.

pkqk (approved):
Looks good @gmac, had a few minor comments but we can merge it after you've responded to them.

gmac:
Committed all suggestions, @pkqk!

pkqk (approved):

PR #117, opened 2021-12-19
Filter out null boundary results before calling nested child step

lucianjon:
## The Problem

When executing a plan with a child step who's parent is itself a child step, there is a unique code path that manipulates the parent step's data and insertion point to make sense to the child.

Anything relying on `data` from [here](https://github.com/movio/bramble/blob/main/query_execution.go#L177) on actually need the non-null boundary results. This [error](https://github.com/movio/bramble/blob/c69e57f7495a624d51261ac893dfec7bbec03032/query_execution.go#L306) in particular will be triggered if the first entry in `data` is null. 

## Reproducing

You need a plan with two levels of child steps, for example:

```
{
  "RootSteps": [
    {
      "ServiceURL": "http://127.0.0.1:61895",
      "ParentType": "Query",
      "SelectionSet": "{ tastyGizmos { id _bramble_id: id } }",
      "InsertionPoint": null,
      "Then": [
        {
          "ServiceURL": "http://127.0.0.1:61896",
          "ParentType": "Gizmo",
          "SelectionSet": "{ wizzle { id _bramble_id: id } _bramble_id: id }",
          "InsertionPoint": [
            "tastyGizmos"
          ],
          "Then": [
            {
              "ServiceURL": "http://127.0.0.1:61897",
              "ParentType": "Wizzle",
              "SelectionSet": "{ bazingaFactor _bramble_id: id }",
              "InsertionPoint": [
                "tastyGizmos",
                "wizzle"
              ],
              "Then": null
            }
          ]
        }
      ]
    }
  ]
}
```

The service `http://127.0.0.1:61896` then needs to return nulls for some or all of its results. The code will then fail when trying to prepare the data for `http://127.0.0.1:61897`. The new test cases cover reproducing.

## The Fix

Simply filter out any null boundary results, they're of no use in the context of executing the child step.

Thanks to @suessflorian for finding the bug and writing the test cases.

### DIFF
--- a/query_execution.go
+++ b/query_execution.go
@@ -172,13 +172,15 @@ func (q *queryExecution) executeChildStep(step *QueryPlanStep, boundaryIDs []str
 
 	q.writeExecutionResult(step, data, nil)
 
-	if len(data) > 0 {
+	nonNillBoundaryResults := extractNonNilBoundaryResults(data)
+
+	if len(nonNillBoundaryResults) > 0 {
 		for _, childStep := range step.Then {
-			boundaryResultInsertionPoint, err := trimInsertionPointForNestedBoundaryStep(data, childStep.InsertionPoint)
+			boundaryResultInsertionPoint, err := trimInsertionPointForNestedBoundaryStep(nonNillBoundaryResults, childStep.InsertionPoint)
 			if err != nil {
 				return err
 			}
-			boundaryIDs, err := extractAndDedupeBoundaryIDs(data, boundaryResultInsertionPoint)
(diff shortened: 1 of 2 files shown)


query_execution.go
```diff
@@ -195,6 +197,18 @@ func (q *queryExecution) executeChildStep(step *QueryPlanStep, boundaryIDs []str
 	return nil
 }
 
+func extractNonNilBoundaryResults(data []interface{}) []interface{} {
```
pkqk, line 200: is this worth extracting as a function?

pkqk (approved):

query_execution.go
lucianjon: I just like having it extracted as it's pretty boilerplatey and I think the calling function reads better with it out.

PR #119, opened 2022-02-09
Improve Bramble timeout support

joshiefu:
## The Issue

Currently, if the limits plugin is used and a service times out, a generic 504 error is returned.

## Reproducing

For example, if you make a query that involves two services, where one is very slow.
```
query {
  serviceA {
    quickField
  }
  serviceB {
    slowField
  }
}
```
You will get  `HTTP/1.1 504 Gateway Timeout` or  `failed to serve query in time`

## The Fix

Instead of using `http.TimeoutHandler`, use the GraphQL client timeout that already exists and is set by the plugin.
This way the same query from above would return the results it has, fills the nullable fields in with nulls and bubble up the errors.

```
{
  "errors": {
    "message": "error during request: Post \"http://localhost:8092/query\": context deadline exceeded",
     ...
  }
  "data": {
    "serviceA": "quick answer",
    "serviceB": null
  }
}
```

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -101,6 +101,26 @@ Set limits for response time and incoming requests size.
 }
 ```
 
+The limit for response time is applied to each sub-query. If a sub-query times out, it will return a null value with a corresponding error denoting the problematic selection set.
+
+e.g.
+
+```json
+{
+  "errors": {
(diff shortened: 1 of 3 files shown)


nmaquet:
@joshiefu PR looks good! Could I just ask you to update the limits plugin docs to make it clear that the timeout is applied to each sub-query?

pkqk (approved):

PR #123, opened 2022-02-17
Handle nullable destinations in results merge of query execution

pkqk:
For fields that are nullable, it is possible that the merge algorithm will encounter some destinations that are `null`. It makes sense to skip those fields.

For example:

```graphql
# Service A
type Gizmo {
    id: ID!
    details: GizmoDetails
}

type GizmoDetails {
    creator: Person!
}

type Person @boundary {
    id: ID!
}

type Query {
    gizmos: [Gizmo!]!
}
````

```graphql
# Service B
type Person @boundary {
    id: ID!
    name: String!
}
type Query {
    person(id: ID!): Person @boundary
}
```

```graphql
query {
    gizmos {
        details {
            creator {
                name
            }
        }
    }
}
```

with results like:
```json
{
    "data": {
        "gizmos": [
            {
                "id": "GIZMO1",
                "details": {
                    "creator": {
                        "name": "Alice"
                    }
                }
            },
            {
                "id": "GIZMO2",
                "details": null
            }
        ]
    }
}
```

The merge algorithm would have expected to be able to walk down the `details`, `creator` path on `GIZMO2`.

### DIFF
--- a/execution_test.go
+++ b/execution_test.go
@@ -2467,6 +2467,72 @@ func TestMergeExecutionResults(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, expected, mergedMap)
 	})
+
+	t.Run("merges with nil destination", func(t *testing.T) {
+		inputMapA := jsonToInterfaceMap(`{
+			"gizmo": {
+				"_bramble_id": "1",
+				"gadgets": [
+					[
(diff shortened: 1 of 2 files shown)


MarinaMovio (approved):

lucianjon (approved):
LGTM!

Issue #112, opened 2021-11-15
Refactor tests that mutate the global log object


lucianjon:
There are some tests that mutate the global log object, using this helper function https://github.com/movio/bramble/blob/main/instrumentation_test.go#L14. Specifically it gives a writer to the log object that is closed at the end of the test. This can cause other tests that run code which logs to print out errors like this: `Failed to write to log, io: read/write on closed pipe`.

pkqk:
The `testing` library now has a [Cleanup](https://pkg.go.dev/testing#T.Cleanup) hook which could be used to restore the original output.

PR #127, opened 2022-02-23
Restore log output after tests finish

joshiefu:
Resolves issue #112 

Using the `Cleanup` hook to restore log to the original output

### DIFF
--- a/instrumentation_test.go
+++ b/instrumentation_test.go
@@ -15,8 +15,15 @@ func collectLogEvent(t *testing.T, f func()) map[string]interface{} {
 	t.Helper()
 	r, w := io.Pipe()
 	defer r.Close()
-	log.SetOutput(w)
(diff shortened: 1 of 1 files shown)


instrumentation_test.go
```diff
@@ -16,6 +16,9 @@ func collectLogEvent(t *testing.T, f func()) map[string]interface{} {
 	r, w := io.Pipe()
 	defer r.Close()
 	log.SetOutput(w)
+	t.Cleanup(func() {
+		log.SetOutput(log.New().Out)
+	})
```
pkqk, line 21: Instead of assuming `log.New().Out` is the correct file I would save the previous value of `Out` in case other redirections have been done.
```golang
prevOut := log.Out
log.SetOutput(w)
t.Cleanup(func() {
  log.SetOutput(prevOut)
})
```

PR #131, opened 2022-03-09
Remove query caching from exposed gateway

LeonHuston:


lucianjon:
LGTM, could you just mention that you removed this too: 

```
	srv.Use(extension.AutomaticPersistedQuery{
		Cache: lru.New(100),
	})
```

Issue #136, opened 2022-03-29
Union of @boundary Types


pasdeloup:
I can't find details about Union support: neither documented nor reported as unsupported.
But I can't make it work:

Graph1
```
type Album @boundary {
  id: ID!
}

type Artist @boundary {
  id: ID!
}

union CatalogEntity = Album | Artist

type FeedCard {
  type: String!
  catalogEntity: CatalogEntity
}

type Query {
  (...)
  feed: [FeedCard]
}
```

Graph2
```
type Album @boundary {
  id: ID!
  title: String!
}

type Artist @boundary {
  id: ID!
  name: String!
}
```

When I query with
```
feed {
    type
    catalogEntity {
      __typename      
      ... on Artist {
        id
        name
      }    	
      ... on Album {
        id
        title
      }
      
    }
  }
}
```

I receive
```
"message": "Cannot query field \"title\" on type \"Artist\"".
```

Did I miss something? maybe declare the Union in graph 2 but how?

lucianjon:
Hi @pasdeloup, at this stage we don't support `@boundary` on union types. It is a little hidden but mentioned [here](https://movio.github.io/bramble/#/federation?id=federation-syntax-faq)

pasdeloup:
Thanks @lucianjon , sorry I really missed that FAQ indeed!

Issue #10, opened 2021-03-08
Allow @boundary enum,  @boundary scalar, and @boundary input


nmaquet:
Shared enums would be really useful to define standard values across services. 
For example:
```graphql
enum Gender @boundary {
    MALE
    FEMALE
    OTHER
    UNKNOWN
}
```

Similarly, shared scalars would be really useful to standardize the encoding of dates and timestamps for example.
For example:
```graphql
scalar ISODate @boundary
```

Shared input types are a bit less useful, but one great use case is to standardize pagination:
```graphql
input PageInput @boundary {
     cursor: ID
     pageSize: Int! = 100
}
```

I think the proper semantics for these new boundary types is that they each appear exactly once in the resulting merged schema. The schema merge operation would fail if the definition of the boundary enums / scalars / inputs differ in any way. This does mean that changing the definition of any of these down the line requires a simultaneous upgrade of all the affected services, but I think that's a price worth paying.

Thoughts @Ackar @pkqk ?

pkqk:
For things like `PageInput` I think the ability to have a shared bit of schema that you could include would help.

It also needs the ability to be defined in multiple services so will need `@boundary` or similar directive.

nmaquet:
Yeah I think reusing `@boundary` is fine here.

pkqk:
Yeah having the same directive is a good idea I think, since the semantics are essentially the same.

There's two things here, we want to be able to have `scalar`s, `enum`s and `input`s with the same type name in various services, and possibly share definitions to reduce duplication of definition and the associated drift.

In the `scalar` case this is easy as it's just a statement that this type name exists.

In the `enum` case we can union all the values from various services into the final enum, although this means that services could get values they are not expecting in their own subset of type system?

Same with `input` types, if we do a union of the `input`s fields then again one service can receive fields it doesn't expect.

It might be safer to have one service that defines the common types we use, mark these with a `@boundary` directive, and then have a way to include this services schema into others as a common package?

nmaquet:
> In the enum case we can union all the values from various services into the final enum, although this means that services could get values they are not expecting in their own subset of type system?

I would only allow `@boundary` union to be defined as the exact same set of values in all services. The alternative causes more issues than is worth dealing with IMHO.

nmaquet:
> Same with input types, if we do a union of the inputs fields then again one service can receive fields it doesn't expect.

I was confused here for a sec, I thought you meant the GraphQL union, not the set union :)

[...]

PR #137, opened 2022-03-30
Improve "not currently supported" part of README.md

pasdeloup:
Following Issue #136 , I just propose a quick README update to make it more explicit that unions can't be shared across services + a link to the FAQ to provide more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,12 @@ It is also stateless and scales very easily.
 
 ## Future work/not currently supported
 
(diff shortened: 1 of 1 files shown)


lucianjon:
Good idea, thanks @pasdeloup. Just to be exhaustive, could you please add there's also no support for shared scalars and enums? There is discussion around this in https://github.com/movio/bramble/issues/10.

lucianjon:
Thanks @pasdeloup

PR #138, opened 2022-03-30
handle failed schema composition in merge test helper

exterm:
Schema composition can fail. If it does, it previously caused a panic through referencing a nil pointer.

In this case, it's better to return early.

I discovered this when experimenting with modifications to the schema composition logic, and instead of failing tests, I got the panic from the test helper.

### DIFF
--- a/merge_fixtures_test.go
+++ b/merge_fixtures_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/vektah/gqlparser/v2"
 	"github.com/vektah/gqlparser/v2/ast"
 )
@@ -89,7 +90,7 @@ func loadAndFormatSchema(input string) string {
 func mustMergeSchemas(t *testing.T, sources ...*ast.Schema) *ast.Schema {
 	t.Helper()
 	s, err := MergeSchemas(sources...)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	return s
 }
 


merge_fixtures_test.go
```diff
@@ -34,6 +34,10 @@ func (f MergeTestFixture) CheckSuccess(t *testing.T) {
 	}
 	actual := mustMergeSchemas(t, schemas...)
 
+	if actual == nil {
+		return
+	}
+
 	// If resulting Query type is empty, remove it from schema to avoid
 	// generating an invalid schema when formatting (empty Query type: `type Query {}`)
 	if actual.Query != nil && len(filterBuiltinFields(actual.Query.Fields)) == 0 {
```
exterm, line 43: This is where the nil pointer dereference happens.

exterm: `mustMergeSchemas` can return `nil`: https://github.com/movio/bramble/blob/70a8e42e99a0c150bf8367dd467270d7d86bf641/merge_fixtures_test.go#L93

lucianjon:
Hi @exterm, good catch and thanks for the PR. So the testing library here has two packages to deal with assertions: `assert` and `require`.  I find it a little odd but if you use an assertion from the `assert` package and the assertional fails, the test will continue on. If you use `require` as your assertion package, it will immediately fail the test. Personally I tend to just use `require` everywhere as it tends to cause less confusion.

In this case given the `must` prefix, I'd expect if the inner call errored out, the calling function wouldn't continue. I think it would be best to just replace `assert.NoError(t, err) ` with `require.NoError(t, err)`.

What do you reckon?

exterm:
>  I'd expect if the inner call errored out, the calling function wouldn't continue. 

Yes, that's what I would have expected! I'm just starting out with go and didn't know the libraries involved. Your proposed solution is clearly better. Will make that change.

lucianjon:
Looks good, thanks @exterm

PR #140, opened 2022-04-06
Use locally built bramble in docker setup

lucianjon:
I wanted to test out some changes I was making and realised our docker-compose setup only uses already published bramble images. Given this is the quickest way to a working setup I figured it's more useful to build the docker image locally.

## Changes

* Changed the main bramble Dockerfile to instead be a self-contained multi-stage docker build, in order to make it easier to work with docker-compose.
* Moved the docker-compose file from examples to the root.
* Updated the github actions to work with the new multi-stage docker build.
* Also migrated the actions to their latest versions.
* Added health checks to each example service so bramble doesn't start before them and immediately exit.
* Ended up having to mess with the linter action as it wasn't happy with my changes.


examples/gqlgen-service/server.go
```diff
@@ -32,6 +33,10 @@ func main() {
 		return next(ctx)
 	}
 	http.Handle("/query", handler.GraphQL(NewExecutableSchema(c)))
+	http.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
```
pkqk, line 37: `http.StatusOK` is sent by default if you start writing to `w`.

I also find it's nicer to use:
```
fmt.Fprintln(w, "OK")
```
so that if you `curl` it from the shell it writes a final newline

docker-compose.yaml
```diff
@@ -0,0 +1,48 @@
+version: "3"
+services:
+  gqlgen-server:
+    build:
+      context: examples/gqlgen-service
+    healthcheck:
```
pkqk, line 6: As far as I know healthchecks don't influence `docker-compose` startup order.
https://docs.docker.com/compose/startup-order/#title see the _tip_ infobox

I see below you've used the `service_healthy` condition, I can find examples of that but there is nothing in the docker-compose yaml documentation for it.

Dockerfile
```diff
@@ -1,6 +1,23 @@
...
+
+COPY go.mod go.sum /workspace/
+
+RUN go mod download
+
+COPY *.go /workspace/
+COPY cmd /workspace/cmd
+COPY plugins /workspace/plugins
+
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="-X 'github.com/movio/bramble.Version=$VERSION'" -o bramble ./cmd/bramble
```
pkqk, line 16: Perhaps don't set the `GOARCH` as this might cause trouble with arm macs etc.

docker-compose.yaml
lucianjon: They don't by default but that means the bramble container dies if you have to build any of the dependent images. 

From what I understand, it was a feature in v2, they removed it in v3 because it caused issues with swarm. They then merged v2 and v3 config formats but kept it undocumented, probably because of swarm again. 

Given it's useful and we have no interest in swarm I'd vote to use it as long as it works.

Dockerfile
lucianjon: As far as I know the image would be `amd64` regardless and arm macs can run `amd64` images. I don't have one though and it doesn't need to be there, so dropped.

examples/gqlgen-service/server.go
lucianjon: Yeah I generally like to explicitly set the header anyway but swapped out for `fmt.Fprintln`

pkqk (approved):

PR #141, opened 2022-04-06
ExecuteQuery refactoring

lucianjon:
## Background

There are problems with `ExecutableSchema.ExecuteQuery` which have been hiding bugs. It's quite a long function that does many things and the fact it's a receiver function on `ExecutableSchema` means it has access to a lot of mutable state it probably shouldn't. Eventually I'd like to entirely redo this function as I don't like the mutex and using shared state approach. I'd rather each instance of a query execution have access to their own copy of the variables. I think extracting out the core execution pipeline from `ExecutableSchema` will make bramble more modular too.

### Bugs Fixed

One of the first things the function does is filter down the schema based on the permissions in the request context. Unfortunately there were plenty of cases further down that just accessed the schema from `ExecutableSchema.MergedSchema` rather than using the filtered schema, essentially bypassing the authorization.

### Changes

* There were a couple instances of redundant checks, like calling `graphql.GetOperationContext(ctx)` as that will actually panic. 
* The extensions were registered right at the end of the function and so if there were any errors you'd never get useful debugging info, that is now at the top.
* Remove errors from `formatResponseData` and `unionAndTrimSelectionSet`. There was only one legitimate error in `formatResponseData` and that was from calling `json.Marshal`. Given this data is already guaranteed to be JSON I decided to panic instead.
* Remove a bunch of functions from `ExecutableSchema` that have no business being on it. There was another bug hiding in `resolveSchema` due to this.


nmaquet (approved):
Excellent. Thanks, @lucianjon.

PR #142, opened 2022-04-07
Reduce HTTP Keep-Alive timeout for HTTP Client

lucianjon:
We've been having issues with services built on nodejs to appear flakey to bramble. There would commonly be errors like `read: connection reset by peer` or `EOF` when trying to poll the service. 

Turns out we were hitting a race condition where our polling period was 5 seconds and the default Keep-Alive for nodejs servers was also 5 seconds. 

We could change the default polling period to something other than 5 seconds but having a short Keep-Alive on the client side seems to be a safe choice. For reference the default on go http clients is 90 seconds.


lucianjon:
Decided to go for just increasing the default polling period.

PR #143, opened 2022-04-07
Increase default polling period to 10 seconds.

lucianjon:
This is to get around a race condition with nodejs based servers who
have a default HTTP Keep-Alive of 5 seconds.


lucianjon:
If we still see issues with the defaults of other servers will have to introduce some kind of random jitter or a very obscure time.

PR #145, opened 2022-04-13
Correctly include subtypes of union in filtered schema.

lucianjon:
When filtering down a schema union sub types could get incorrectly
excluded from the final schema.

This occurs when the permissions do not specify AllowAll on the union
type but rather use field level permissions.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -180,7 +180,7 @@ func filterDefinition(sourceSchema *ast.Schema, visited map[string]bool, types m
 				// Node interface is not defined in the merged schema
 				continue
 			}
-			if typ.Kind == ast.Interface {
+			if typ.IsAbstractType() {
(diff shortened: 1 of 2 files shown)


auth.go
```diff
@@ -194,12 +194,6 @@ func filterDefinition(sourceSchema *ast.Schema, visited map[string]bool, types m
 			_ = filterDefinition(sourceSchema, visited, types, sourceSchema.Types[typeName], AllowedFields{AllowAll: true})
 		}
 
-		// unions
```
pkqk, line 197: What was the reasoning for this step being here originally?

pkqk (approved):

auth.go
lucianjon: From what I could gather, it was a special case to deal with unions but didn't function correctly because it ignored the union's possible types.

PR #152, opened 2022-04-26
Add WrapGraphQLClientTransport to Plugin API

nmaquet:
This PR adds a `WrapGraphQLClientTransport` method to the Plugin API. This allows plugins to hook into the outbound GraphQL request flow of Bramble to perform tasks such as tracing / logging / etc.

This PR also adds a `InterceptRequest` method to the Plugin API. This allows plugins to get notified each time Bramble serves a GraphQL query.

### DIFF
--- a/client.go
+++ b/client.go
@@ -26,7 +26,7 @@ type GraphQLClient struct {
 type ClientOpt func(*GraphQLClient)
 
 // NewClient creates a new GraphQLClient from the given options.
-func NewClient(opts ...ClientOpt) *GraphQLClient {
+func NewClient(plugins []Plugin, opts ...ClientOpt) *GraphQLClient {
 	c := &GraphQLClient{
 		HTTPClient: &http.Client{
 			Timeout: 5 * time.Second,
@@ -38,6 +38,9 @@ func NewClient(opts ...ClientOpt) *GraphQLClient {
 		opt(c)
 	}
 
+	for _, plugin := range plugins {
+		c.HTTPClient.Transport = plugin.WrapGraphQLClientTransport(c.HTTPClient.Transport)
+	}
 	return c
 }
 
--- a/client.go
+++ b/client.go
@@ -26,7 +26,11 @@ type GraphQLClient struct {
 type ClientOpt func(*GraphQLClient)
 
 // NewClient creates a new GraphQLClient from the given options.
-func NewClient(plugins []Plugin, opts ...ClientOpt) *GraphQLClient {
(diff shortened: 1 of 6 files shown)


client.go
```diff
@@ -26,7 +26,7 @@ type GraphQLClient struct {
 type ClientOpt func(*GraphQLClient)
 
 // NewClient creates a new GraphQLClient from the given options.
-func NewClient(opts ...ClientOpt) *GraphQLClient {
+func NewClient(plugins []Plugin, opts ...ClientOpt) *GraphQLClient {
```
LeonHuston, line 29: Do we care about this being a breaking change for current `NewClient` consumers?

LeonHuston (approved):
A small comment around the exposed API and can you add some docs for this change as well :)

client.go
pkqk: @nmaquet I added [a1e312a](https://github.com/movio/bramble/pull/152/commits/a1e312a1e65e0f89fae200d6422e3e1e9f206b4b) which makes the plugin constructor a new function instead of changing the `NewClient` signature.

nmaquet: Perfect, thanks.

plugin.go
```diff
@@ -49,6 +53,11 @@ func (p *BasePlugin) GraphqlQueryPath() (bool, string) {
 	return false, ""
 }
 
+// InterceptRequest is called before bramble starts executing a request.
+// It can be used to inspect the unmarshalled GraphQL request bramble receives.
+func (p *BasePlugin) InterceptRequest(ctx context.Context, operationName, rawQuery string, variables map[string]interface{}) {
```
sodabrew, line 58: I might have suggested naming this something like `WrapClientRequest` because it's on the external-client-side of a Bramble instance.

plugin.go
```diff
@@ -59,6 +68,11 @@ func (p *BasePlugin) ApplyMiddlewarePrivateMux(h http.Handler) http.Handler {
 	return h
 }
 
+// WrapGraphQLClientTransport wraps the http.RoundTripper used for GraphQL requests.
+func (p *BasePlugin) WrapGraphQLClientTransport(transport http.RoundTripper) http.RoundTripper {
```
sodabrew, line 72: I might have suggested naming this something like `WrapOriginTransport` because it's on the origin-server-side of a Bramble instance.

sodabrew (commented):
Comments on naming. Suggest establishing a naming convention for client-side and origin-side, wrap vs. middleware, mux vs. request vs. transport, etc.

Issue #151, opened 2022-04-25
Fragment not working as expected for interface types


azadasanali:
**Problem Statement**

When an interface type is implemented in graphql schema, we are able to query with spread operator and `on Type` pattern.
But the same cannot be achieved by using Fragments as exemplified below 

**Example Request**
```
interface Book {
  title: String!
}

type Textbook implements Book {
  title: String! # Must be present
  author: String!
}

type Magazine implements Book {
  title: String! # Must be present
  publisher: String!
}

type Query {
  books: [Book!] # Can include Textbook objects
}

fragment TextBookFragment on Textbook {
      author # Only present in Textbook
}

fragment MagazineFragment on Magazine {
      publisher # Only present in Magazine
}

# this fails
query GetBooks {
  books {
    __typename
    title
    ... TextBookFragment
    ... MagazineFragment
  }
}

# this passes
query GetBooks {
  books {
    __typename
    title
    ... on Textbook {
     author
    }
    ... on Magazine {
      publisher
    }
  }
}

```
Response if TextBook is not found in federated query we are getting
```
{
  "errors": [
    {
      "message": "got a null response for non-nullable field \"author\"",
    }
}
```

lucianjon:
Thanks @azadasanali, I've managed to reproduce it and we'll get onto fixing it.

azadasanali:
@lucianjon When this will be released?

lucianjon:
@azadasanali We’ll get a release out when we’re  back in the office on Monday

nmaquet:
@azadasanali Here you go! https://github.com/movio/bramble/releases/tag/v1.4.0

PR #153, opened 2022-04-26
Fix fragment spread elimination during result merge + format

lucianjon:
FIxes: #151 

There was an issue multiple top level named fragment spreads under an interface type were not being correctly eliminated during the merge and format stages of query execution.

The behaviour should be that for both inline and named fragment spreads, only the spread that matches the concrete type in the response is kept. This was working for inline spreads but not named spreads.

The problem was made a little harder to fix due to an existing problem with `evaluateSkipAndInclude` that would overwrite the `ObjectDefinition` for a `FragmentSpread` with the wrong value.

This PR fixes both issues.

### DIFF
--- a/execution_test.go
+++ b/execution_test.go
@@ -984,6 +984,59 @@ func TestFederatedQueryFragmentSpreads(t *testing.T) {
 		f.checkSuccess(t)
 	})
 
+	t.Run("with multiple top level fragment spreads (gadget implementation)", func(t *testing.T) {
+		f := &queryExecutionFixture{
+			services: []testService{serviceA, serviceB},
+			query: `
+			query Foo {
+				snapshot(id: "GADGET1") {
+					id
+					name
+					... GadgetFragment
+					... GizmoFragment
+				}
+			}
+
+			fragment GadgetFragment on GadgetImplementation {
+				gadgets {
+					id
+					name
+					agents {
+						name
+						... on Agent {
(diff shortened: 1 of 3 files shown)


execution_test.go
```diff
@@ -984,6 +984,59 @@ func TestFederatedQueryFragmentSpreads(t *testing.T) {
...
+							country
+						}
+					}
+				}
+			}
+
+			fragment GizmoFragment on GizmoImplementation {
+				gizmos {
+					id
+					name
```
pkqk, line 1016: should we have a test for an alias on a field too?

pkqk (approved):

execution_test.go
lucianjon: The code that deals with aliases is slightly unrelated to fragments as aliases are defined on an `*ast.Field` and here we're testing `*ast.InlineFragment` and `*ast.FragmentSpread`. There are other tests covering correct handling of aliases.

dilip9292: Hi @lucianjon 
 in the test case If we fetch either Gadjet Or Gizmo it works. If we want to fetch a list of object where object could be either Gadject or Gizmo it would not work. Can you suggest what can we do in such case

azadasanali: @dilip9292 I think if gadget and gizmo have common interface this will not work

PR #159, opened 2022-06-16
Split large files into smaller chunks

lucianjon:
`execution.go` and `execution_test.go` were large enough where they would sometimes be unwieldy to work with.

This PR keeps `execution.go` for the tests that run the entire execution pipeline, adds `execution_introspection_test.go` for the introspection tests that go against the internal resolver, adds `execution_helpers_test.go` for the tests that run against a particular function in the execution pipeline. Then `execution_result.go` and `execution_result_test.go` for the functions that operate on the results post execution.

Couple other small style tidy ups around the place too.


pkqk:
@lucianjon if you can resolve the conflicts on `execution.go` we can get this in.

PR #163, opened 2022-07-20
Use correct spelling in limits plugin

pkqk:
The `MaxResponseTime` field was misspelled as `MaxResponeTime`.


nmaquet (approved):
Well the json tag was correctly spelled, at least.

PR #165, opened 2022-07-27
Populate PossibleTypes with merged type

pkqk:
The `mergePossibleTypes` function when populating the types just used the value from one of the source trees instead of the merged type.

This would only be apparent once one type of an `interface` was a `@boundary` type, depending on the order the services were merged in the final merged schema could be missing fields from other services.

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -242,7 +242,7 @@ func mergePossibleTypes(sources []*ast.Schema, mergedTypes map[string]*ast.Defin
 				for _, possibleType := range possibleTypes {
 					if possibleType.Name != nodeInterfaceName {
 						if ast.DefinitionList(result[typeName]).ForName(possibleType.Name) == nil {
-							result[typeName] = append(result[typeName], possibleType)
+							result[typeName] = append(result[typeName], mergedTypes[possibleType.Name])
 						}
 					}
 				}
--- a/merge.go
+++ b/merge.go
@@ -234,15 +234,15 @@ func mergeDirectives(sources []*ast.Schema) map[string]*ast.DirectiveDefinition 
 func mergePossibleTypes(sources []*ast.Schema, mergedTypes map[string]*ast.Definition) map[string][]*ast.Definition {
 	result := map[string][]*ast.Definition{}
 	for _, schema := range sources {
-		for typeName, interfaces := range schema.PossibleTypes {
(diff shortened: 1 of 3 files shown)


merge.go
```diff
@@ -234,15 +234,15 @@ func mergeDirectives(sources []*ast.Schema) map[string]*ast.DirectiveDefinition
...
 					continue
 				}
-				for _, i := range interfaces {
-					if i.Name != nodeInterfaceName {
-						if ast.DefinitionList(result[typeName]).ForName(i.Name) == nil {
-							result[typeName] = append(result[typeName], i)
+				for _, pt := range pts {
+					if pt.Name != nodeInterfaceName {
+						if ast.DefinitionList(result[typeName]).ForName(pt.Name) == nil {
+							result[typeName] = append(result[typeName], mergedTypes[pt.Name])
```
pkqk, line 245: This change here from `pt` to `mergedTypes[pt.Name]` is the actual fix for the issue.

merge_test.go
```diff
@@ -923,3 +923,79 @@ func TestMergeWithAlternateId(t *testing.T) {
...
+				id: ID!
+				foo: Boolean!
+			}
+
+			type Bar implements Face {
+				id: ID!
+				bar: Boolean!
+			}
+
+            type Query  {
```
lucianjon, line 946: Just some inconsistent indentation here.

merge.go
```diff
@@ -234,15 +234,15 @@ func mergeDirectives(sources []*ast.Schema) map[string]*ast.DirectiveDefinition
...
+		for typeName, pts := range schema.PossibleTypes {
 			if typeName != serviceObjectName && typeName != nodeInterfaceName {
 				if _, ok := mergedTypes[typeName]; !ok {
 					continue
 				}
-				for _, i := range interfaces {
-					if i.Name != nodeInterfaceName {
-						if ast.DefinitionList(result[typeName]).ForName(i.Name) == nil {
-							result[typeName] = append(result[typeName], i)
+				for _, pt := range pts {
```
lucianjon, line 242: Could we rename `pt` -> `possibleType` and `pts` -> `possibleTypes`. Don't mind the shorter names in loop variables in general but it is a bit of indirection in this case to get from `pt` -> `schema.PossibleTypes`.

lucianjon (approved):
Nice find

PR #166, opened 2022-08-24
Abstract type boundary federation

mai00cmi:
I think there are two ways to fix this
a) a886f69b5fc7cd0dd2ab924e8477d21b61bb9d6a just assume if the _bramble_id is missing that it's not a boundary type
b) 2dcec5a23e7b3366c0bbc533ee90c2818f46b612 always ask for __typename and check the types before merging.

I'm not sure which one is the correct one.

### DIFF
--- a/execution.go
+++ b/execution.go
@@ -387,9 +387,6 @@ func extractBoundaryIDs(data interface{}, insertionPoint []string) ([]string, er
 		switch ptr := ptr.(type) {
 		case map[string]interface{}:
 			id, err := boundaryIDFromMap(ptr)
-			if err != nil {
-				return []string{}, nil
-			}
 			return []string{id}, err
 		case []interface{}:
 			result := []string{}
--- a/execution.go
+++ b/execution.go
@@ -387,6 +387,9 @@ func extractBoundaryIDs(data interface{}, insertionPoint []string) ([]string, er
 		switch ptr := ptr.(type) {
 		case map[string]interface{}:
 			id, err := boundaryIDFromMap(ptr)
+			if err != nil {
+				return []string{}, nil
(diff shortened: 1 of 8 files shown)


pkqk:
We have the type information, so checking that makes more sense to me rather than just making assumptions.

nmaquet:
Thanks for spending some time on this @mai00cmi! Great to see you digging into Bramble 🤘

pkqk:
@mai00cmi if we squash and remove the error checks in [fe1681f](https://github.com/movio/bramble/pull/166/commits/fe1681f88bcd4e895f5f2fc580b710793da591f1) and [a886f69](https://github.com/movio/bramble/pull/166/commits/a886f69b5fc7cd0dd2ab924e8477d21b61bb9d6a) does your final commit solve the issue?

Happy to merge after that if that solves the error.

execution.go
```diff
@@ -375,23 +374,32 @@ func extractAndDedupeBoundaryIDs(data interface{}, insertionPoint []string) ([]s
 		deduped = append(deduped, id)
 	}
 
-	return sort.StringSlice(deduped), nil
```
pkqk, line 378: Do you think the sort is unnecessary or is it part of the fix? I'd prefer to leave other behaviour changes to a separate PR if possible

mai00cmi: That's not a function call. It's an unnecessary type cast `type StringSlice []string`. It was never sorting. Do you want it sorted?

pkqk: ah I see, no if it's unnecessary then remove it, it was possibly left over from a refactor.

pkqk (approved):

PR #167, opened 2022-08-29
Pass variables downstream

mai00cmi:
Instead of inlining variables into downstream queries, which failed when the serialisation of the value from json didn't match the graphql syntax (eg if the variable was a complex json type) filter the variables map to the values required by the downstream query.

This should enable the definition of things like a json scalar value without bramble interfering and generating a syntactically invalid query.

### DIFF
--- a/client.go
+++ b/client.go
@@ -159,9 +159,9 @@ type Request struct {
 }
 
 // NewRequest creates a new GraphQL requests from the provided body.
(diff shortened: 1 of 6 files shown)


pkqk (approved):
I've rebased this after @andyyu2004's changes since the operation name in query generation conflicted.

PR #174, opened 2022-10-12
Propogate `operationName` to subqueries

andyyu2004:
This is useful for tracing and metrics. Currently child queries are made with an empty `operationName`.

### DIFF
--- a/execution.go
+++ b/execution.go
@@ -291,14 +291,16 @@ func (q *queryExecution) createGQLErrors(step *QueryPlanStep, err error) gqlerro
 // have insertionPoint: ["foo", "bar", "movies", "movie", "compTitles"], with the below example as the boundary result we're
 // crawling for ids:
 // [
-// 	 {
-//     "_bramble_id": "MOVIE1",
-//     "compTitles": [
-//       {
-//   	   "_bramble_id": "1"
-// 		 }
-//	   ]
-//   }
+//
+//		 {
+//	    "_bramble_id": "MOVIE1",
+//	    "compTitles": [
+//	      {
+//	  	   "_bramble_id": "1"
+//			 }
+//		   ]
+//	  }
+//
 // ]
 //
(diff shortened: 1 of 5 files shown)


execution.go
```diff
@@ -291,14 +294,16 @@ func (q *queryExecution) createGQLErrors(step *QueryPlanStep, err error) gqlerro
 // have insertionPoint: ["foo", "bar", "movies", "movie", "compTitles"], with the below example as the boundary result we're
 // crawling for ids:
 // [
-// 	 {
```
pkqk, line 294: can you remove the indentation/whitespace changes?

execution.go
```diff
@@ -126,7 +128,8 @@ func (q *queryExecution) executeRootStep(step *QueryPlanStep) error {
 
 func (q *queryExecution) executeDocument(document string, serviceURL string, response interface{}) error {
 	req := NewRequest(document).
-		WithHeaders(GetOutgoingRequestHeadersFromContext(q.ctx))
+		WithHeaders(GetOutgoingRequestHeadersFromContext(q.ctx)).
+		WithOperationName(q.operationName)
```
pkqk, line 132: I think you'll need to add the operation name to the generated query document too, otherwise the downstream queries will fail as it won't be able to find the operation in the query?

pkqk: I think in the `buildBoundaryQueryDocuments` and `executeRootStep` functions

execution.go
andyyu2004: I'm getting the changes just from running `gofmt`. Do you mind if they stay?

pkqk: `gofmt` really shouldn't be touching comments, but can you pull them out into a different commit then?

execution.go
andyyu2004: I've added the `operationName` to the required documents. I haven't added any automated tests but have checked it fixes the issue for the example setup.

PR #188, opened 2023-01-13
Fixed key in "config.json.example" to match source code (config.go)

anar-khalilov:
image


pkqk (approved):
Nice catch, thank you.

Issue #204, opened 2023-05-01
Add Datadog graphql trace


marcosuma:
Hi there (again) :), our infra uses Datadog to monitor our services. I recently noticed that for our bramble-based service we are using an [_http tracer_](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http) but this has the advantage that it'll not correctly log request errors (graphql always returns 200). 

So I noticed that the same Datadog library supports [**graph-gophers**](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/contrib/graph-gophers/graphql-go) - see the example tab for how to use it - which would likely be ideal for us in order to get proper graphql monitoring.

The problem is that I am not sure this could be bundled into our bramble-based service. Could a plugin be used? I tried with the Init method but no luck.

nmaquet:
Hey @marcosuma! Yes, you should be able to write a plugin for your use case. At Movio, we use AWS X-Ray for tracing so we've developed a plugin for it. I've created a gist for you to have a look at the code: https://gist.github.com/nmaquet/0c04308d7ae7c234387926cf4c7e791c

As you can see from the X-Ray plugin, most of the heavy lifting happens in `Init`, `WrapGraphQLClientTransport` and `InterceptRequest`. I'm quite confident you should be able to use a very similar approach for a Datadog plugin.

marcosuma:
Thanks @nmaquet! That helps a lot. Is there a way I can get access to [GraphQL errors](https://pkg.go.dev/github.com/movio/bramble@v1.4.3#GraphqlError) in a plugin? What I would basically like to do is to trace those errors and publish metrics about it. Current situation is that if I capture the body response with a `http.RoundTripper` whenever there is an error the response.body is actually empty (here's a [gist](https://gist.github.com/marcosuma/456b3244465be51c4cf271f579f4e3b1#file-gistfile1-txt) about it)

nmaquet:
Hey @marcosuma. If you want easy access to GraphQL errors, the `Plugin` interface needs to be extended a bit. I've done so here: https://github.com/movio/bramble/pull/205 Can you give this a shot and we'll try and merge this and release it soon.

marcosuma:
That's awesome! I can try to clone it and test it locally, but are you plannign to release a new version?

nmaquet:
[...]

PR #205, opened 2023-05-02
Add InterceptResponse to the Plugin interface

nmaquet:
InterceptResponse is called after bramble has finished executing a request. It can be used to inspect and/or modify the response bramble will return.

Added to help with https://github.com/movio/bramble/issues/204

### DIFF
--- a/executable_schema.go
+++ b/executable_schema.go
@@ -167,7 +167,7 @@ func (s *ExecutableSchema) ExecuteQuery(ctx context.Context) *graphql.Response {
 	})
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


LeonHuston (approved):
Looks fine to me, maybe want to add an example of some ways you can use this (maybe doing what that other person was asking about)

pkqk (approved):
Agree with @LeonHuston comment about examples, we should probably have one for each plugin method.

Changes look good.

pkqk (commented):
Can you squash the typo commit though @nmaquet

nmaquet:
> Can you squash the typo commit though @nmaquet

Done!

PR #207, opened 2023-05-04
Only return Fields, Types for Services with Schema

pkqk:
Unreachable services haven't loaded a schema yet so the definition is nil, causing the meta plugins query to fail with a nil dereference panic.


nmaquet:
I think it would be slightly nicer to add 
```
if schema == nil {
    return nil
}
```
to both `getTypes` and `getFields` instead of changing the `Services` method? It's of course equivalent but it will avoid others making the same mistake in the future and keeps the loop in `Services()` tidier.

nmaquet (approved):
:shipit:

PR #211, opened 2023-05-30
Split error update metrics into gauge and counter

pkqk:
Use counter for tracking number of failures and the gauge for current error states.

We were finding once the error state was triggered it could not be reset as the counter is intended to be used as an ever incrementing metric.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -15,10 +15,20 @@ var (
 		Help: "A gauge representing the current status of remote services schemas",
 	})
 
-	promServiceUpdateError = prometheus.NewCounterVec(
+	promServiceUpdateErrorCounter = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
+			Name: "service_update_error_total",
+			Help: "A counter indicating how many times services have failed to update",
+		},
+		[]string{
+			"service",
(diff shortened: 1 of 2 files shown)


metrics.go
```diff
@@ -15,10 +15,20 @@ var (
 		Help: "A gauge representing the current status of remote services schemas",
 	})
 
-	promServiceUpdateError = prometheus.NewCounterVec(
+	promServiceUpdateErrorCounter = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
+			Name: "service_update_failures",
```
lucianjon, line 20: Was already incorrect but counters should generally have a `_total` suffix.

metrics.go
```diff
@@ -15,10 +15,20 @@ var (
...
+			Help: "A counter indicating how many times services have failed to update",
+		},
+		[]string{
+			"service",
+		},
+	)
+
+	promServiceUpdateError = prometheus.NewGaugeVec(
```
lucianjon, line 28: Would rename `s/promServiceUpdateError/promServiceUpdateErrorGauge` so it's the same as the counter.

PR #216, opened 2023-06-20
Handle empty fragment spread being first in query

pkqk:
The previous fix only handled cases where an empty fragment returned nothing after a field had already been written, if it was first and returned nothing the subsequent fields would write an invalid comma ahead of the `innerBody`.

As discussed in #215 we need to rewrite this in a more robust way without relying on string building but this should handle the bug for now.


andyyu2004 (approved):
Yep, this will do the trick.

Issue #96, opened 2021-11-07
Support for `Node` interface-based federation.


nmaquet:
Services implementing [Relay's Global Object Identification Spec](https://relay.dev/graphql/objectidentification.htm) (i.e. the `Node` interface should be allowed to be configured as boundary types without the need for an additional boundary accessor field.

giautm:
Hey @nmaquet, I'm interested in this issue. So, is there any way to support/workaround for it?

TheBeachMaster:
Hi, is there a proposed work-around for this? Anyone using [ent](https://entgo.io/) will run into this issue.

nmaquet:
@TheBeachMaster there no current plans to implement this but if someone is willing to give this a shot, I'd be happy to guide the efforts and assist with the PR.

PR #217, opened 2023-07-23
fix: :alien: Added support for nodes query

karatekaneen:
Ent (sometimes?) generates a "nodes" query to fetch multiple nodes in batch in the latest version . This was not compatible with Bramble so I fixed that.

The queries that I get generated in my project looks like this:

```graphql
type Query {
  """Fetches an object given its ID."""
  node(
    """ID of the object."""
    id: ID!
  ): Node
  """Lookup nodes by a list of IDs."""
  nodes(
    """The list of node IDs."""
    ids: [ID!]!
  ): [Node]!
}
```

Maybe related to #96, not sure


pkqk:
Hi @karatekaneen is this the [ent](https://entgo.io/docs/graphql) you mean? Can you post an example of generating the graphql `nodes` field.

The support for the Relay Node interface is relatively unused at the moment, it's left over from an initial prototype design of how our federation model would work.

karatekaneen:
@pkqk Yes, it's Ent with `entgql` I'm referring to. The problem was that when I was trying to add my service to Bramble I got an error with something like "unknown type Node" which I wanted to get rid off.

My Ent schema is currently something of a kitchensink but it looks like this:

```go
package schema

import (
	"entgo.io/contrib/entgql"
	"entgo.io/contrib/entproto"
	"entgo.io/ent"
	"entgo.io/ent/dialect/entsql"
	"entgo.io/ent/schema"
	"entgo.io/ent/schema/field"
)

type FormData struct {
	ent.Schema
}

func (FormData) Annotations() []schema.Annotation {
	return []schema.Annotation{
		entsql.Annotation{Table: "data"},
		entgql.RelayConnection(),
		entgql.Directives(entgql.Directive{Name: "boundary"}),
		entgql.QueryField(),
		entproto.Message(),
		entproto.Service(entproto.Methods(entproto.MethodGet)),
	}
}

// Fields of the FormData.
func (FormData) Fields() []ent.Field {
	return []ent.Field{
		field.Int("object_id").
			StorageKey("obj_id").
			Annotations(entproto.Field(2)),

		field.String("module").
			StorageKey("data_mod").
			Annotations(entproto.Field(3)),

		field.Int("type").
			StorageKey("data_type").
			Annotations(entproto.Field(4)),
	}
}

// Edges of the FormData.
func (FormData) Edges() []ent.Edge {
	return []ent.Edge{}
}
```

And the relevant parts of GraphQL that gets generated looks like this:

```graphql
type FormData implements Node @boundary {
  id: ID!
  objectID: Int!
  module: String!
  type: Int!
}

interface Node @goModel(model: "bitbucket.org/company/entdemo/ent.Noder") {
  id: ID!
}

type Query {
  node(id: ID!): Node
  nodes(ids: [ID!]!): [Node]!
  formDataSlice(
    after: Cursor
    first: Int
    before: Cursor
    last: Int
    where: FormDataWhereInput
  ): FormDataConnection!
}
```

So, just to be clear. I don't mind the Relay Node interface to be unused. It's just that I wanted to be able to hook up my projects to Bramble without having to filter out the incompatible query from every schema in the `service` query.

karatekaneen:
@pkqk Had any chance to take a look at this?

pkqk:
Hi @karatekaneen I haven't had a chance to review it yet, though your use case makes more sense now. I might be able to find time end of next week. Are you ok running your fork in the mean time?

karatekaneen:
@pkqk It's ok to run the fork for now. Just wanted to avoid syncing it when you are doing updates.

karatekaneen:
@pkqk Found a small bug where the custom http.Client wasn't being respected. Fixed that on this branch as well

pkqk:
Thanks @karatekaneen, would you be able to pull it out into it's own PR? I can merge and test that more quickly. We haven't had a chance to test with Ent yet.

PR #225, opened 2023-08-31
Don't exit on misconfigured plugins

andyyu2004:
I don't think a misconfigured plugin should not prevent bramble from starting up. The handling should be analogous to the case where the plugin is unregistered, warn and move on.

### DIFF
--- a/config.go
+++ b/config.go
@@ -245,9 +245,11 @@ func (c *Config) ConfigurePlugins() []Plugin {
 			log.Warnf("plugin %q not found", pl.Name)
 			continue
 		}
+
 		err := p.Configure(c, pl.Config)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -245,9 +245,11 @@ func (c *Config) ConfigurePlugins() []Plugin {
 			log.Warnf("plugin %q not found", pl.Name)
 			continue
 		}
+
 		err := p.Configure(c, pl.Config)
 		if err != nil {
-			log.WithError(err).Fatalf("error unmarshalling config for plugin %q: %s", pl.Name, err)
+			log.WithError(err).Warnf("error unmarshalling config for plugin %q (skipping plugin)", pl.Name)
```
andyyu2004, line 251: Also I dropped the `err` from the format string. It's in logrus context anyway.

PR #226, opened 2023-09-27
Correct doc comment in mergeMaps

pkqk:
- the comment was wrong
- two panic messages were identical in different paths, making it harder to identify which path was failing


andyyu2004 (commented):
Yep that would have been very useful when debugging. I would also add the actual type that was provided to the panic.

lucianjon (approved):

Issue #228, opened 2023-11-13
Long running request to service 


codedge:
Hey!

I experience a weird behaviour when having long running requests in my service connected to the gateway.
I run a PHP-based service behind the gateway, that sometimes needs up to 45s to return the response. In 90% of the cases the response is not returned by the gateway and instead I get a `Socket hang up` back.

I already enabled the _Limits_ plugin and put this there 

```json
{
      "name": "limits",
      "config": {
        "max-response-time": "120s",
        "max-request-bytes": 1000000
      }
}
```

This removes the initial `reached timeout` error message, but still I have the problem that the connection between the gateway and the service somehow gets lost.

I am 100% sure, that the response is correct and is returned by (backend) service.

Does that somehow sound familiar to you or any hint where to look?

Thanks!

pkqk:
Hi @codedge, can you post a copy of the response you're getting? The string `Socket hang up` seem to show up when I search the go stdlib and you mention later it's a `reached timeout`.

Does it log the request when it fails?

codedge:
Sorry for the confusion.

**1. Resolving the `reached timeout`**

At first I got a `reached timeout` error. This error was directly visible inside the logs of Bramble. I figured out, that by using the _limits_ plugin with the above mentioned configuration, I can get around this error.

This is solved ✔️ 

**2. The `Socket hang up` problem**

This error is returned by `curl` (or Postman) or any other GraphQL client. There is no other response. 

**Error in `curl`**

![2023-11-13_224641](https://github.com/movio/bramble/assets/4409904/882f2676-197f-4661-9c6c-abf00d870e71)

**Error in Postman**

![2023-11-13_225222](https://github.com/movio/bramble/assets/4409904/dd5cff7c-9888-4992-a66d-5a24adc5af7f)

I tend to say this is some `keepAlive`/`idle` timeout problem.

I also found [this link](https://forum.golangbridge.org/t/empty-reply-from-server-for-long-running-task-in-http-handler/26430), which talks about the `net.http.Server.WriteTimeout`.

codedge:
.. and I can confirm, that changing the [`WriteTimeout`](https://github.com/movio/bramble/blob/main/main.go#L61) to f. ex. `60`

```golang
func runHandler(ctx context.Context, wg *sync.WaitGroup, name, addr string, handler http.Handler) {
	srv := &http.Server{
		Addr:         addr,
		Handler:      handler,
		ReadTimeout:  5 * time.Second,
[...]

PR #229, opened 2023-11-15
Add configuration options for net/http timeouts

pkqk:
These timeouts should be tuneable for users who don't want to limit downstream services to only accept fast responses.

Address #228 by providing the option to increase the write timeout.

@codedge can you give this a review and see if it addresses your issue, it wasn't able to be configured by the `limits` plugin as it's the `main` method that sets up the `http.Server` instances.

In the `config.json` you can either provide a `default-timeouts` map of `read`, `write` and `idle` keys that will be used for all the `http.Server` instances or override the defaults with `gateway-timeouts` and `private-timeouts` maps that only need to provide values differing from the defaults.

### DIFF
--- a/config.go
+++ b/config.go
@@ -21,19 +21,31 @@ type PluginConfig struct {
 	Config json.RawMessage
 }
 
+type TimeoutConfig struct {
+	ReadTimeout          string        `json:"read"`
+	ReadTimeoutDuration  time.Duration `json:"-"`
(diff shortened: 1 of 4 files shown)


codedge:
Going to check during the day and will let you know.

Thanks already for the fast response!

codedge:
This works perfectly for me. 👍 Can be merged!

PR #234, opened 2023-11-21
Prevent panic when Service struct is nil

pkqk:
If a service is unreachable the `Services` map could be unpopulated, so fetching the service name is not possible.


andyyu2004 (approved):
Is there a more useful value to provide than the empty string for when we go inside the first branch, but `svc` is nil? e.g. `unreachable`. Otherwise looks good.

pkqk:
It already has `Unreachable` in the `status` field so I think blank is better than using the field for multiple pieces of info in an unstructured way.
```json
{
    "name": "",
    "version": "",
    "serviceUrl": "https://…",
    "status": "Unreachable",
    "schema": "",
    "__typename": "BrambleService"
},
```

PR #235, opened 2023-11-22
Fetch schema updates in parallel

Quantumplation:
When running bramble in a Lambda, the startup time is critical. This fetches the schema updates in parallel, which dramatically reduces the startup time.

### DIFF
--- a/executable_schema.go
+++ b/executable_schema.go
@@ -79,31 +79,44 @@ func (s *ExecutableSchema) UpdateSchema(forceRebuild bool) error {
 		}
 	}()
 
-	for url, s := range s.Services {
-		logger := log.WithField("url", url)
-		updated, err := s.Update()
-		if err != nil {
-			promServiceUpdateErrorCounter.WithLabelValues(s.ServiceURL).Inc()
-			promServiceUpdateErrorGauge.WithLabelValues(s.ServiceURL).Set(1)
-			invalidSchema, forceRebuild = true, true
-			logger.WithError(err).Error("unable to update service")
-			// Ignore this service in this update
-			continue
-		}
-		promServiceUpdateErrorGauge.WithLabelValues(s.ServiceURL).Set(0)
-		logger = log.WithFields(log.Fields{
-			"version": s.Version,
-			"service": s.Name,
-		})
+	var wg sync.WaitGroup
+	var mutex sync.Mutex
 
-		if updated {
-			logger.Info("service was updated")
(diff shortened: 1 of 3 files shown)


pkqk (commented):
Thanks @Quantumplation 

I'll give it a test before merging it

executable_schema.go
```diff
@@ -79,32 +79,45 @@ func (s *ExecutableSchema) UpdateSchema(forceRebuild bool) error {
...
-
-		if updated {
-			logger.Info("service was updated")
-			updatedServices = append(updatedServices, s.Name)
-		}
+	var wg sync.WaitGroup
+	var mutex sync.Mutex
+
+	for url_, s_ := range s.Services {
+		wg.Add(1)
```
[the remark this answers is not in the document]
Quantumplation: Great suggestion, updated

pkqk:
@Quantumplation one other point, depending on the number of services being federated a loop like that might have other adverse performance implications. Perhaps a limit on the number of parallel requests at a time is a good idea.

Quantumplation:
Usually a rate limit at 64, and I had a hard time imagining having 64 services but I can add that no problem

executable_schema.go
```diff
@@ -79,32 +80,49 @@ func (s *ExecutableSchema) UpdateSchema(forceRebuild bool) error {
...
-			"version": s.Version,
-			"service": s.Name,
-		})
-
-		if updated {
-			logger.Info("service was updated")
-			updatedServices = append(updatedServices, s.Name)
-		}
+	// Only fetch at most 64 services in parallel
+	rateLimit := make(chan struct{}, 64)
```
pkqk, line 84: Would [`errgroup.SetLimit`](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit) provide the same behaviour as the `rateLimit` channel?

Quantumplation: Technically it would be slightly different, as the call to `errgroup.Go` would block on the main thread, instead of the green thread itself blocking, but I don't think that makes a material difference, so I can use that if you prefer.

pkqk: I think it's clearer to read and it shouldn't matter which goroutine blocks, since the `errgroup.Wait` call blocks afterwards anyway

Quantumplation: Yea agreed; apparently this method was only added in September of 2022, hence why I wasn't aware of it 😅 Certainly a lot more convenient!

PR #246, opened 2023-12-20
feat: OpenTelemetry tracing and metrics

asger-noer:
## Telemetry

This PR adds build-in support for OpenTelemetry. All GraphQL queries to the federated services served from bramble are instrumented, this includes service schema updates, service list updates, as well as federated queries from users. The telemetry is exported using the gRPC exporter.

### Configuration

Since telemetry is build into bramble the configuration also lives in the bramble root `config.json`. 

`enabled` will enable telemetry.
`url` are the url for the collector other system capable of receiving OTLP metrics and spans.
`insecure` allows skipping the TLS check when communicating with the collector.
`service_name` allows for the span service name to be set. Useful in multi-deployment settings.

## `GraphQLClient` changes

The GraphQL client has been changed so it now returns the raw timeout error if such error is received. This allows for easier error type checking the processing the error.

## Testing

Modified the `TestQueryExecutionServiceTimeout` to check on the type instead of an error string. This required some modifications to the run function, that now take a function that will test the response, moving the actual testing back into the overall test function itself. Nice.


main.go
```diff
@@ -28,6 +30,18 @@ func Main() {
...
 	go cfg.Watch()
 
+	shutdown, err := InitTelemetry(context.TODO(), cfg.Telemetry)
+	if err != nil {
+		log.WithError(err).Error("error creating telemetry")
+	}
+
+	defer func() {
+		log.Info("flushing and shutting down telemetry")
+		if err := shutdown(context.TODO()); err != nil {
```
pkqk, line 40:
```suggestion
		if err := shutdown(ctx); err != nil {
```
does the Backgound context work here instead of using TODO?

main.go
```diff
@@ -28,6 +30,18 @@ func Main() {
 	}
 	go cfg.Watch()
 
+	shutdown, err := InitTelemetry(context.TODO(), cfg.Telemetry)
```
pkqk, line 33:
```suggestion
	shutdown, err := InitTelemetry(ctx, cfg.Telemetry)
```

main.go
```diff
@@ -51,6 +65,7 @@ func Main() {
 	go runHandler(ctx, &wg, "public", cfg.GatewayAddress(), cfg.GatewayTimeouts, gtw.Router(cfg))
 
 	wg.Wait()
+
```
pkqk, line 68:
```suggestion
```

introspection.go
```diff
@@ -31,9 +31,10 @@ func NewService(serviceURL string, opts ...ClientOpt) *Service {
 }
 
 // Update queries the service's schema, name and version and updates its status.
-func (s *Service) Update() (bool, error) {
-	req := NewRequest("query brambleServicePoll { service { name, version, schema} }").
-		WithOperationName("brambleServicePoll")
+func (s *Service) Update(ctx context.Context) (bool, error) {
+	req := NewRequest("query brambleServicePoll { service { name, version, schema} }")
+	req.WithOperationName("brambleServicePoll")
```
pkqk, line 36: This is working because the `With...` Functions are modifying the request which is a pointer but the intention is for them to return a copy of the request, so I think we should leave it as chained method calls.

executable_schema.go
```diff
@@ -155,6 +180,21 @@ func (s *ExecutableSchema) ExecuteQuery(ctx context.Context) *graphql.Response {
 	operation := operationCtx.Operation
 	variables := operationCtx.Variables
 
+	ctx, span := s.tracer.Start(ctx, "Executable Schema Query",
+		trace.WithSpanKind(trace.SpanKindInternal),
+	)
+
+	defer span.End()
+
+	handleErr := func(err error) {
```
pkqk, line 189: perhaps give this a less generic name and say what it is doing? `traceError`, `recordError` or `addErrorToSpan`

client.go
```diff
@@ -93,15 +110,35 @@ func WithUserAgent(userAgent string) ClientOpt {
...
+		trace.WithAttributes(
+			attribute.String("graphql.operation", request.OperationName),
+			attribute.String("graphql.query", request.Query),
+			attribute.String("graphql.variables", fmt.Sprintf("%v", request.Variables)),
+		),
+	)
+
+	defer span.End()
+
+	handleErr := func(err error) error {
```
pkqk, line 124: Same naming question as the other usage of this pattern below

schema.go
```diff
@@ -6,6 +6,18 @@ import (
 	"github.com/vektah/gqlparser/v2/ast"
 )
 
+// Operation is a GraphQL operation type.
+type Operation string
```
pkqk, line 10: I don't see this type offering much benefit, as it is always cast to a string before use and should always be coming from a valid graphql query. The swithc in `WithOperationType` could handle that case without the need for the type wrapper

asger-noer: I think this type could be useful for replacing `ParentType` in the struct below. But that requires a bigger change, that I think would be nice to have in another PR. What do you think about this? 😄

```go
// plan.go

// QueryPlanStep is a single execution step
type QueryPlanStep struct {
	ServiceURL     string
	ServiceName    string
	ParentType     string --> Operation
	SelectionSet   ast.SelectionSet
	InsertionPoint []string
	Then           []*QueryPlanStep
}
```

pkqk: If we're going to make that change I'd suggest making it all in it's own PR instead of a partial change in this one.

asger-noer: I removed it for now 😄

.gitignore
```diff
@@ -4,3 +4,5 @@
 .idea
 .DS_Store
 coverage.out
+*__debug_*
+config.json
```
pkqk, line 8: Can you set the file ending?

client.go
```diff
@@ -177,7 +219,23 @@ func (r *Request) WithHeaders(headers http.Header) *Request {
 	return r
 }
 
+func (r *Request) WithOperationType(operation string) *Request {
+	switch strings.ToLower(operation) {
+	case "query":
```
pkqk, line 224: perhaps compress it to one case? I was surprised you can't do an assignment in a switch statement though
```golang
op := strings.ToLower(operation)
switch op {
case "query", "mutation", "subscription":
  r.OperationType = op
default:
  r.OperationType = "query"
}
```

execution.go
```diff
@@ -258,7 +269,8 @@ func (q *queryExecution) createGQLErrors(step *QueryPlanStep, err error) gqlerro
 
 	var gqlErr GraphqlErrors
 	var outputErrs gqlerror.List
-	if errors.As(err, &gqlErr) {
+	switch true {
```
pkqk, line 272:
```suggestion
	switch {
```
You don't need the true if you're using the case statement as an if/else?

format_test.go
```diff
@@ -396,7 +396,7 @@ func TestFormatDocument(t *testing.T) {
 		string(operationDefinition.Operation),
 		operationDefinition.SelectionSet,
 	)
-	assert.Equal(t, `query {    search(id: "123") {        id        title    } }`, res)
+	assert.Equal(t, "query {\n  search(id: \"123\") {\n    id\n    title\n  }\n}", res)
```
pkqk, line 399: We could make a comparison function that compared a query ignoring whitespace? Then we don't need to tweak the tests when we change formatting?

pkqk (commented):
Just a couple if small things, then we can squash the commits.

.gitignore
asger-noer: It is the executable build when debugging, so it haven't got any extension.

format_test.go
asger-noer: Yes, I agree. The whitespace in these queries are all over the place. I think this would be nice to do in another PR for now.

asger-noer: Yes, I agree. The whitespace in these queries are all over the place. I think this would be nice to do in another PR for now.

Issue #254, opened 2024-02-26
Input types missing default values


benzolium:
When I (or frontend lib) query merged schema:

```
{
  __type(name:"someName"){
    kind
    name
    inputFields {
      defaultValue
      name
    }
  }
}
```

I'm getting null as default values, tho I actually have some:
```
{
  "data": {
    "__type": {
      "kind": "INPUT_OBJECT",
      "name": "someName",
      "inputFields": [
        {
          "defaultValue": null,
          "name": "name"
        },
        {
          "defaultValue": null,
          "name": "description"
        },
      ]
    }
  }
}
```

I've reproduced it in a test: https://github.com/movio/bramble/commit/a7c3cbeaf10c3a166a9dad720704ee8ade63d433.
I want to fix it, but I don't know, where to start. @pkqk, maybe you could guide me?
I've already checked gqlparser - it looks fine. 
I have also written a test in `merge_test.go` and it's passing, but I have doubts because `CheckSuccess` helper is a bit complicated.

andyyu2004:
Hey @benzolium, relevant code is in `resolveField` and generally `resolve*` functions in `executable_schema.go`. Let me know if you need some further guidance on the implementation.

benzolium:
Thanks @andyyu2004! I might be confused with `"defaultValue"` case in `resolveInputValue` so I've overlooked it.
PR with a fix is ready.

PR #256, opened 2024-02-27
Provide defaultValue on input types in schema introspection

benzolium:
Resolves https://github.com/movio/bramble/issues/254

### DIFF
--- a/execution_introspection_test.go
+++ b/execution_introspection_test.go
@@ -53,6 +53,9 @@ func TestIntrospectionQuery(t *testing.T) {
 
 	input CinemaInput {
 		name: String! = "SomeName"
+		cinema: CinemaInput = {name: "SomeOtherName"}
+		mainGenre: MovieGenre = ACTION
+		description: String
 	}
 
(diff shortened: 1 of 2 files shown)


andyyu2004 (commented):
Thanks! I'd like to see a few more test cases with more types of the defaultValue (enums, input objects etc).

benzolium:
@andyyu2004 sure thing! Got it.

execution_introspection_test.go
```diff
@@ -53,6 +53,9 @@ func TestIntrospectionQuery(t *testing.T) {
 
 	input CinemaInput {
 		name: String! = "SomeName"
+		cinema: CinemaInput = {name: "SomeOtherName"}
```
pkqk, line 56: is this a valid input type? Would it create a infinite chain of `cinema` values.

benzolium: Yes indeed:
https://spec.graphql.org/October2021/#sec-Input-Objects.Circular-References

PR #263, opened 2024-04-19
Inject execution debugging info into query plan

lucianjon:
As each step is completed, inject some information into the step that aids in debugging step failures and timing.

The information is accessed on the result object when the below request header is sent: `X-Bramble-Debug: plan`

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -31,17 +31,15 @@ type executionStepResult struct {
 }
 
 func (e *executionStepResult) MarshalJSON() ([]byte, error) {
-	type toMarshal struct {
+	return json.Marshal(&struct {
 		Executed  bool
 		Error     error `json:",omitempty"`
 		TimeTaken string
-	}
-	j := toMarshal{
+	}{
 		Executed:  e.executed,
 		TimeTaken: e.timeTaken.String(),
 		Error:     e.error,
-	}
-	return json.Marshal(&j)
+	})
 }
 
 // MarshalJSON marshals the step the JSON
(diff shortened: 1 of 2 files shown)


plan.go
```diff
@@ -19,26 +20,60 @@ type QueryPlanStep struct {
...
 		Variables: map[string]interface{}{},
 	})
-	return json.Marshal(&struct {
-		ServiceURL     string
-		ParentType     string
-		SelectionSet   string
-		InsertionPoint []string
-		Then           []*QueryPlanStep
-	}{
+	type toMarshal struct {
```
pkqk, line 56: Is it worth extracting these out to locally named types instead of the anonymous structs?
The only conditional assignment is based on if they are nil values so the `omitempty` tag on the field will handle that.

pkqk (commented):
Maybe this should be switched on based on `timings` being provided in the `X-Bramble-Debug` header

lucianjon:
> Maybe this should be switched on based on `timings` being provided in the `X-Bramble-Debug` header

Yeah I considered that but I thought it was a bit unexpected injecting the entire plan when asking for `timings`.

plan.go
lucianjon: They are named types, just scoped to the function. I don't see a need to move them out if they're only used for json marshalling. As to why they're there, that pattern was already used and I do like keeping all of that information unexported and only exposing it as we marshal.

pkqk: Oh I didn't mean to extract them, I meant is there any benefit to giving them local names inside the function when they were anonymous before. i.e go back to
```golang
return json.Marshal(&struct{
	ServiceURL          string
	ParentType          string
	SelectionSet        string
	InsertionPoint      []string
	ExecutionStepResult *executionStepResult `json:",omitempty"`
	Then                []*QueryPlanStep
}{
	ServiceURL:          s.ServiceURL,
	ParentType:          s.ParentType,
	SelectionSet:        formatSelectionSetSingleLine(ctx, nil, s.SelectionSet),
	InsertionPoint:      s.InsertionPoint,
	Then:                s.Then,
	ExecutionStepResult: s.executionResult,
})
```

lucianjon: Yeah right, no reason. I've changed it back to anonymous structs.

Issue #154, opened 2022-04-28
Request syntax error while trying to upload file


timspin:
Hi!

I've implemented file uploading using these library https://github.com/jaydenseric/graphql-multipart-request-spec for my graphql service. And it works correctly when I querying my service directly. But in case when there is bramble between I get error like:
`input:1: Syntax Error GraphQL request (1:36) Expected Name, found String`

So my question is: does bramble support those kind of queries with multipart/form-data content type?

nmaquet:
Hi @timspin. This isn't something that we have used / found a need for yet but it does sound interesting. As you've noticed, Bramble doesn't currently support this but we would welcome any contributions.

pkqk:
It looks like adding the [`MultipartForm`](https://github.com/99designs/gqlgen/blob/master/graphql/handler/transport/http_form.go) transport to [`gateway.go`](https://github.com/movio/bramble/blob/main/gateway.go#L39) will be needed

benzolium:
It also looks like one will need to set `multipart/form-data` for downstream services [here](https://github.com/movio/bramble/blob/9a383e6d8ee90cd6f4708df8441e9972dc42e569/client.go#L109).

benzolium:
@nmaquet @pkqk could you please guide me where to start implementing this feature?

I think I'd have to do something like [this in nautilus](https://github.com/nautilus/gateway/blob/deb7b43ac849154303d90835b349cb642981a3a0/http.go#L251), but in bramble.

pkqk:
@benzolium the `github.com/99designs/gqlgen/graphql/handler` module should handle the receiving of multipart requests, I haven't tested it but I think the uploaded file streams will be available as variables.

You've identified the right spot in `client.go` where you'll need to add the option to make a multipart request to the downstream service in [your comment above](https://github.com/movio/bramble/issues/154#issuecomment-1473962456)

You could add something to the context that indicates if the incoming request used a multipart request, and then decide if based on the variables present (they are filtered down to only those necessary for the downstream request) if the downstream request needs to be a multipart request. (I wouldn't do it by default as it's not a standard transport and perhaps not all downstream services support it)

You might have to construct the multipart request yourself based on the spec, I haven't found a module that does it for you.

gmac:
FWIW, I added this feature to Ruby stitching recently: https://github.com/gmac/graphql-stitching-ruby/pull/103. There’s some documentation and code there that may serve as a motivational example of the feature.

PR #268, opened 2024-05-25
Add multipart/form-data transport to support file uploads

benzolium:
Resolves https://github.com/movio/bramble/issues/154

@pkqk this a draft. Please take a look. Any suggestions or remarks?

I'm currently testing it manually. 
Though I'm using almost the same approach for 6 month on production. 🙂


benzolium:
There is a problem with logging.
My code:
https://github.com/movio/bramble/blob/3dc62ad5a8ef4af565af56459bce13ae3d943f36/instrumentation.go#L58
breaks tests in `instrumentation_test.go`:
However, code from the head:
https://github.com/movio/bramble/blob/949badb3574efe87f16a1bdde6a27a0136d1babd/instrumentation.go#L55 
just logs the whole requests body with log level info. Which is kinda not ok in case of large files (it even caused me some OOMs).

pkqk:
Hi @benzolium,

This looks cool, I'll give it a test when I have a chance. Is it possible you could add an example server to the `examples` folder to show it in use?

benzolium:
@pkqk sure. Done.

I've also added one more test case and updated docs with `headers` plugin config: to make file upload work we **must** pass `Content-Type` headers, so I've updated `gateway.json` config.

benzolium:
There’s still a problem with logging the whole body. For large files it will be a big issue. 

@pkqk i cannot just remove logging of body completely because there is some instrumentation logic relying on that (there are even tests for having body in logs). So what should I do? Remove body from logging by checking content type header?

pkqk:
It's probably worth removing the request body from the logging by default, it was in there initially as it was useful during early development.

For now you could stop it logging the body on a `multipart/form-data` request with something like:
```diff
diff --git a/middleware.go b/middleware.go
index a370351..f74f874 100644
--- a/middleware.go
+++ b/middleware.go
@@ -104,15 +104,20 @@ func addRequestBody(e *event, r *http.Request, buf bytes.Buffer) {
 	contentType := r.Header.Get("Content-Type")
 	e.addField("request.content-type", contentType)
 
-	if r.Method != http.MethodHead &&
-		r.Method != http.MethodGet &&
-		contentType == "application/json" {
-		var payload interface{}
-		if err := json.Unmarshal(buf.Bytes(), &payload); err == nil {
-			e.addField("request.body", &payload)
-		} else {
+	if r.Method != http.MethodHead && r.Method != http.MethodGet {
+		switch contentType {
+		case "application/json":
+			var payload interface{}
+			if err := json.Unmarshal(buf.Bytes(), &payload); err == nil {
+				e.addField("request.body", &payload)
+			} else {
+				e.addField("request.body", buf.String())
+				e.addField("request.error", err)
+			}
+		case "multipart/form-data":
+			e.addField("request.body", fmt.Sprintf("%d bytes", len(buf.Bytes())))
+		default:
 			e.addField("request.body", buf.String())
-			e.addField("request.error", err)
 		}
 	} else {
 		e.addField("request.body", buf.String())
```

client.go
```diff
@@ -132,8 +134,23 @@ func (c *GraphQLClient) Request(ctx context.Context, url string, request *Reques
...
 
 	var buf bytes.Buffer
-	if err := json.NewEncoder(&buf).Encode(request); err != nil {
-		return traceErr(fmt.Errorf("unable to encode request body: %w", err))
+	contentType := "application/json; charset=utf-8"
+	requestContentType := ""
+	if request.Headers != nil {
+		requestContentType = request.Headers.Get("Content-Type")
+	}
+	if !strings.Contains(requestContentType, "multipart") {
```
pkqk, line 142:
```suggestion
	if requestContentType == "multipart/form-data" {
```
How about inverting this if statement, so it treats it as multipart if it's certain it can, and have everything else as the default case.

I don't think you need to check for other `multipart` mime types as the spec only specifies behaviour for `multipart/form-data`

benzolium: @pkqk sure, I'll inverse the statement. However, we should probably still use `contains` as `Content-Type` header will also have boundary in case of multipart:
```
"request.content-type":"multipart/form-data; boundary=--------------------------208172004141002940919311"
```

benzolium: I've inverted the statement, changed `Contains` to `HasPrefix` and specified `multipart/form-data`.

pkqk:
@benzolium I tried testing the example service but I'm getting these errors
```json
{
  "errors": [
    {
      "message": "map[string]interface {} is not an Upload",
      "path": [
        "uploadGizmoFile",
        "upload"
      ],
      "locations": [
        {
          "line": 2,
          "column": 2
        },
        {
          "line": 3,
          "column": 2
        }
      ],
      "extensions": {
        "selectionPath": null,
        "selectionSet": "{ uploadGizmoFile(upload: $fileA) uploadGadgetFile(upload: {upload:$fileB}) }",
        "serviceName": "gqlgen-service",
        "serviceUrl": "http://gqlgen-multipart-file-upload-service:8080/query"
      }
    },
    {
      "message": "map[string]interface {} is not an Upload",
      "path": [
        "uploadGadgetFile",
        "upload",
        "upload"
      ],
      "locations": [
        {
          "line": 2,
          "column": 2
        },
        {
          "line": 3,
          "column": 2
        }
      ],
      "extensions": {
        "selectionPath": null,
        "selectionSet": "{ uploadGizmoFile(upload: $fileA) uploadGadgetFile(upload: {upload:$fileB}) }",
        "serviceName": "gqlgen-service",
        "serviceUrl": "http://gqlgen-multipart-file-upload-service:8080/query"
      }
    }
  ],
  "data": {
    "uploadGizmoFile": null,
    "uploadGadgetFile": null
  }
}
```

This is the curl command I'm using
```sh
curl --url http://localhost:8082/query \
  --form 'operations={"query":"mutation upload($fileA: Upload!, $fileB: Upload!) {\n\tuploadGizmoFile(upload: $fileA)\n\tuploadGadgetFile(upload: {\n\t\tupload: $fileB\n\t})\n}","variables":{"fileA":null,"fileB":null}}' \
  --form 'map={
 "a": ["variables.fileA"],
 "b": ["variables.fileB"]
}' \
  --form a=@a.txt \
  --form b=@b.txt
```

benzolium:
@pkqk good catch. Fixed. I was populating files map incorrectly.

client.go
```diff
@@ -368,20 +368,43 @@ func prepareMultipartData(request *Request) (*prepareMultipartDataResult, error)
...
-	fw.Write(input)
+	_, err = fw.Write(input)
+	if err != nil {
+		return nil, err
+	}
 	fw, err = mpw.CreateFormField("map")
 	if err != nil {
 		return nil, err
 	}
+	_, err = fw.Write([]byte("{"))
```
pkqk: instead of manually writing the json here you could use:
```golang
json.NewEncoder(fw).Encode(res.fileMap)
```
to write the already existing map to the stream. The slice in the value is only ever of length one as it's never added to.

client.go
```diff
@@ -275,3 +292,122 @@ func (e GraphqlErrors) Error() string {
...
+		if ioErr != nil {
+			return nil, ioErr
+		}
+	}
+	err = mpw.Close()
+	if err != nil {
+		return nil, err
+	}
+	contentType := mpw.FormDataContentType()
+	return &prepareMultipartDataResult{
```
pkqk, line 409: I don't think you're gaining much here with the named struct in the return, you could change the return value of the function to be the `(bytes.Buffer, string, error)` instead?

client.go
```diff
@@ -275,3 +292,122 @@ func (e GraphqlErrors) Error() string {
...
+	res := parseMultipartVariables(request.Variables)
+
+	var fw io.Writer
+	var buf bytes.Buffer
+	mpw := multipart.NewWriter(&buf)
+	fw, err := mpw.CreateFormField("operations")
+	if err != nil {
+		return nil, err
+	}
+	input, err := json.Marshal(request)
```
pkqk, line 367: you can write the `Marshal` output directly with
```golang
if err = json.NewEncoder(fw).Encode(request); err != nil {
```

pkqk:
@benzolium I had a bunch more suggestions that were a bit too big to do through the review system. If you'd like to [look at my fork](https://github.com/benzolium/bramble/compare/multipart-file-upload...pkqk:multipart-suggestions?expand=1)

Also looking through the spec examples, I noticed the current file map approach won't handle [arrays of Files](https://github.com/jaydenseric/graphql-multipart-request-spec?tab=readme-ov-file#file-list), i.e making a map with `variables.fileList.1`, `variables.fileList.2` etc.

benzolium:
@pkqk thanks for suggestions! I'd definitely apply some of them. 

However there are several things I'd like to discuss:
- `mime.ParseMediaType` instead of just `HasPrefix` (mime is doing much more than checking prefix and content type always starts with type/subtype as of [RFC](https://www.rfc-editor.org/rfc/rfc1521.html#page-9))
- combining `case graphql.Upload, *graphql.Upload` removes some duplicated code, however it requires costly type assertion (which we are already doing)
- `isMultipart` is deciding if downstream request should be a multipart one, but:
    - user has already decided that request is multipart with content-type
    - I cannot imagine the case where it is possible for user to sent a multipart request and some downstream service should receive post request
    - `isMultipart` iterates variables one more time which is kind of overhead

Should we open a PR to my fork and discuss suggestions there? Or maybe I should apply other suggestions and we could discuss remaining here?

As for array files, I'll look into that. Haven't ever used files array so overlooked.

pkqk:
Hi @benzolium, good questions, performance wise, we are dealing with a lot of json parsing and serialisation which is going to be an order of magnitude slower than any of the checks we do here, so I wouldn't consider optimising these until it proves to be an issue.

- parsing the media type using the mime library gives us a few more robustness checks, such as normalising the typename, we could do it with `strings.HasPrefix("type/subtype", strings.ToLower(strings.TrimSpace(contentType)))` but I think using the mime library is clearer.
- I'm not sure of the overhead of the type assertion, but I would expect the compiler to be able to skip it as it's already done it once and the variable value hasn't changed, so I think an optimiser would be able to unroll the case statement.
- The multipart request type isn't part of the base graphql transport, so it's possible another service doesn't support it, so if the request was being federated to two services, one that supported multipart and one that didn't then it could break.
- iterating over the map twice is more overhead, but O(n) + O(n) is still O(n), only with a small increase in the constant factor.

In general I'd go for clarity and slightly less efficient code than optimising before it's necessary.

pkqk:
@benzolium I can either PR into your branch, or you can cherry pick the ones that you agree with and discuss the other ones here.

benzolium:
Hey @pkqk! Thanks for the answers and for the ping. You can surely PR into my branch. 

Sorry for the delay. I’m a little but overwhelmed with work now.

pkqk:
There's no rush @benzolium, looking forward to merging this in but I want us to get it polished and working well before we do.

benzolium:
Hey @pkqk! I've merged your suggestions and added a case for list of files.

RomBurunduk:
I have a problem with isMultipart
If in a query one of the fields is an array of structure, where one of the fields is Upload, then this Upload is not found.

```
input SomeInput {
    documents: [DocInput!]
}

input DocInput {
    file: Upload
    fileName: String
}
```

In this case, the ‘documents’ is []interface{} and does not go on the stack for further processing.

PR #271, opened 2024-06-13
Stop bubble up of error on nullable list item

benzolium:
Fix case when there is a nullable null value in array item:
```graphql
type Gizmo {
	id: ID!
	color: String!
}
type Query {
	gizmos: [Gizmo]!
}
```
```json
{
  "gizmos": [
	  { "id": "GIZMO1", "color": "RED" },
	  null,
	  { "id": "GIZMO2", "color": "GREEN" }
  ]
}
```


pkqk:
@benzolium I rebased your PR as your other pr had created a conflict with `go.sum`

Issue #80, opened 2021-10-29
RFC: execution batching


gmac:
I'm curious where execution batching could fit within Bramble's roadmap, and if the feature as a whole aligns with the green path for the tool. Bear with me please for a slightly contrived example... I've previously [framed this scenario in a video](https://www.youtube.com/watch?v=VmK0KBHTcWs&t=263s) if you'd like to skip the written introduction. We have two simple schemas:

```graphql
# -- shop schema

type Shop {
   products: [Product]!
}
type Product @boundary {
  id: ID!
}
type Query {
  shop1: Shop
  product(id: ID): Product @boundary # << useless accessor
}

# -- product schema

type Product @boundary {
  id: ID!
  name: String!
}
type Query {
  products(ids: [ID!]!): [Product]! @boundary
}
```

Then we have this query:

```graphql
query {
  shop1 {
    products {
      name
    }
  }
  shopA: shop1 {
    products {
      name
    }
  }
}
```

This contrived example demonstrates an inefficiency: we've initiated two _separate_ requests to the products service during the same resolver generation for the two `products` fields. This is traditionally avoidable using [batch execution](https://www.graphql-tools.com/docs/batch-execution), which consolidates multiple operations created during the same execution frame into one outbound request, such as:

```graphql
query($_0_ids: [ID!]!, $_1_ids: [ID!]!) {
  _0_products: products(ids: $_0_ids)
  _1_products: products(ids: $_1_ids)
}
```

This combined operation issues a single network request; which guarantees the operations arrive at the destination together and get multiplexed; which in turn maximizes the opportunities for database batching. Correctly implemented, this service would only perform one database lookup for both `products` accessors thanks to batching.

All that's to say – is there anything prohibitive in the Bramble architecture that should prevent this from working? The pattern simply wraps each Service client in a dataloader that will merge and unpack the individual requests made to it. 

I don't mean this as a feature request so much as a discussion point for an optimization roadmap. Is this a feature that would harmonize with Bramble's goals in the long term? We built this into GraphQL Tools, though we stole it [from Gatsby](https://github.com/gatsbyjs/gatsby/pull/22347), and it works great. Rah, rah, open source!

gmac:
Closing the loop on this one – @nmaquet indicated this is a fine roadmap feature, although not a priority of core development. Open for contributions. I may take a pass as a Go learning challenge, though have no commitments on a timeline. In the meantime, we can probably close this one out.

benzolium:
[...]

PR #272, opened 2024-06-13
Allow adding transports before defaults

benzolium:
As I mentioned in https://github.com/movio/bramble/issues/80#issuecomment-2137465210, one could not override default transports now.

With this small change, one could set up own transport, because first transport get's selected:
https://github.com/99designs/gqlgen/blob/fb67b709af4f0c5ba6eff8222c728b076a6eb09b/graphql/handler/server.go#L98

It will allow users, for example, to introduce batching with a bramble plugin, as I also mentioned in comment: https://github.com/movio/bramble/issues/80#issuecomment-2137465210.


pkqk (approved):
This looks good, I couldn't find any plugins that depend on the order for that hook.

PR #283, opened 2024-10-17
Include operation type and name in log message

pkqk:
This is intended to help us differentiate requests in the logs based on the operation name. With custom filters we could use the fields on the event but automatic tools like DataDog appear to group on the `msg` key.

So a request will now log something like:
```json
{
  "duration": "11.3045ms",
  "level": "info",
  "msg": "query:randomGizmo",
  "operation.name": "randomGizmo",
  "operation.type": "query",
  "request.body": {
    "operationName": "randomGizmo",
    "query": "query randomGizmo($d: String!) {\n  a: randomGizmo {\n    id\n    delay(duration: $d)\n\t\trating\n\t\temail\n\t\tname\n  }\n  b: randomGizmo {\n    id\n    delay(duration: $d)\n\t\trating\n\t\temail\n\t\tname\n  }\n}",
    "variables": {
      "d": "5ms"
    }
  },
  "request.content-type": "application/json",
  "request.path": "/query",
  "response.size": 229,
  "response.status": 200,
  "time": "2024-10-17T01:51:36.70727097Z"
}
```

### DIFF
--- a/instrumentation.go
+++ b/instrumentation.go
@@ -11,7 +11,7 @@ import (
 const eventKey contextKey = "instrumentation"
 
 type event struct {
-	name      string
+	nameFunc  EventNameFunc
 	timestamp time.Time
 	fields    EventFields
 	fieldLock sync.Mutex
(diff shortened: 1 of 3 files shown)


facecrusher (approved):
LGTM

PR #295, opened 2025-01-28
Package bumps

LeonHuston:
We have left the packages alone for a little bit so this is just here to bump some of them, trims the dep tree a bit which is quite nice


telemetry.go
```diff
@@ -13,7 +13,7 @@ import (
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 	"go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
-	semconv "go.opentelemetry.io/otel/semconv/v1.25.0"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
```
pkqk, line 16: @LeonHuston is this the package that needs to be kept in sync elsewhere? I remember it causing services to not start up?

LeonHuston: Yep, thats why theres that small test in `telemtry_test.go` which fails if you revert this diff

